PR #6, opened 2022-01-23
Checked and added info about IO APIC.

DeanoBurrito:
Added TipsAndTricks file.
Added info about hardware breakpoints.
Added info about "memory" clobber.
Expanded section on RSDT, added mirror info for XSDT.


DeanoBurrito:
Shouldn't conflict with your existing PR on the IOREDTBL info :)

PR #7, opened 2022-01-30
Keyboard notes

dreamos82:


Keyboard.md
```diff
@@ -0,0 +1,25 @@
+# Keyboard Driver
+
+## IRQ and IO-Apic
+
+* The Keyboard IRQ is the number 1. This is mapped to the IO Redirection Table 1, at offset 12h and 13h (called IOREDTBL1).
```
DeanoBurrito, line 5: this is a good example, but perhaps its worth being more general when describing this:
Irq1 will be pin 1 on the IO APIC, which is controlled by entry 1 in the redirect table. This register is accessed as 2x 32bit registers, where the register number is as follows: 0x10 + (irq_number * 2). In this case that means the lower 32 bits of irq 1 are in register 0x12, with the upper 32 bits in 0x13.

Keyboard.md
```diff
@@ -0,0 +1,25 @@
...
+## Handling Keyboard interrupt
+
+* When a key is pressed the scancode is made available on the keyboard buffer, and the IRQ handler is supposed to read it, if it is not done this prevent subsequents IRQs from the keyboard.
+* The scancodes on the buffer can be read using IN/OUT assembly instructions at the port 0x60. The code will look like to something similar: 
+```C
+    int scancode = inb(0x60);
+    do_something_with_the_scancode(scancode);
+```
+
+* An irq is raised when a key is pressed, and when it is released too. The bit #7 is 0 when the key is pressed, an 1 otherwise.
```
DeanoBurrito, line 18: this is only true if using scancode set 1. This actually gets surprisingly messy as the default for ps/2 keyboard is scancode set 2 (different encoding), but the ps2 controller (which is what you're talking to on ports 0x60 + 0x64, not the keyboard directly) defaults to translating set 2 scancodes to set 1 scancodes. However if your bios initializes the keyboard, and for some reason removes the ps2 controller's translation without restoring it, you might end up with junk data. 
Supporting set 1 is fine, but be sure to check what scancode set the keyboard is using, and check the ps2 controller is translating to set 1 if needed (or you can just reset both devices, as this is the behaviour on a reset).

PR #22, opened 2022-05-02
Keyboard info section

dreamos82:


Add_Keyboard_Support/DriverImplementation.md
```diff
@@ -0,0 +1,217 @@
+# Handling scancodes
+
+Handling the keyboard interrupt was pretty simple, we just basically needed to add an irq to the APIC and read the scancode. But at the end what we got is just a number, and we want to eventually translated it into a character. 
+
+In this section we will see see how to build the Keyboard driver, and what are its responsibilities. 
+
+First of all the driver is not responsible of translating the scancode into printable characters (well sometime on early stages of development it does it too...)
```
DeanoBurrito, line 7: yes! Although I would say its responsible for translating keyboard specific scancodes (set 1/2/3) into some vendor neutral format that any program can use.

Add_Keyboard_Support/DriverImplementation.md
```diff
@@ -0,0 +1,217 @@
...
+
+What it cares is only to keep track of the various keyboard events and make them available to any service that needs them.
+
+As already mentioned there are 3 different scancode sets, we will focus on just one (the set 1, since most of the keyboard even if using a different set will have the controller that automatically translates the scancode to that), by the way we will try to implement a generic way to translate, so when eventually a new set needs to be added the changes needed will be very little.
```
DeanoBurrito, line 11: the set 1 -> set 1
"will have the controller" what controller? the keyboard controller? I would say "will have the ps2 controller automatically translate set 2 (the default on power-on) to set 1."

Add_Keyboard_Support/DriverImplementation.md
```diff
@@ -0,0 +1,217 @@
+# Handling scancodes
+
+Handling the keyboard interrupt was pretty simple, we just basically needed to add an irq to the APIC and read the scancode. But at the end what we got is just a number, and we want to eventually translated it into a character.
```
DeanoBurrito, line 3: Perhaps this could rephrased a little. The ps2 controller is a system device, so it's irq line was enabled in the ioapic, then an entry was added in the idt. Now inside of the interrupt handler we just added we can read the scancode from the keyboard.

Add_Keyboard_Support/DriverImplementation.md
```diff
@@ -0,0 +1,217 @@
...
+* Handle the caps, num, screen locks (with the leds)
+* We need to translate the scancode into a human readable character when needed
+
+From now on we will assume that the scancode translation is enabled, so no matter what set is being used it will be translate to set 1. 
+
+## High Level Overview
+
+In the previous section we have seen how the interrupt was generated and how to read data from the keyboard. But now we need to write a proper driver, trying to address the issues listed above (well not all of them since some are an higher level than they will be implemented "using" the driver, not by it.
+
+We will try to build the driver in small step adding one piece at time, so it will be easier to understand it.
```
DeanoBurrito, line 28: step -> steps

Add_Keyboard_Support/DriverImplementation.md
```diff
@@ -0,0 +1,217 @@
...
+```
+
+If we want to store just the scancode we don't need much more so we can already implement our new irq handling function:
+
+```c
+
+void keyboard_driver_irq_handler() {
+
+    
+    int scancode = inb(0x60); // Read byte from the Keyboard data port
```
DeanoBurrito, line 50: Maybe a `uint8_t` is more appropriate than an `int` here?

Add_Keyboard_Support/DriverImplementation.md
```diff
@@ -0,0 +1,217 @@
...
+```c
+#define NORMAL_STATE 0
+#define PREFIX_STATE 1
+
+uint8_t current_state;
+```
+
+Now there are three scancodes thate are composed by  4 or 5, we are not going to cover them, but it can be a good exercise trying to find states for them too. The keys are Print Screen pressed/released and Pause (only pressed, it doesn't have a release state).
+
+Now the first question is: do we need to store the prefix? Well for this one the answer is that it mostly depends on the design decision for the driver, but it is not necessary, especially if we are going to translate the keyboard scancodes into kernel scancodes (in this way we don't store a set specific code, and supporting multiple scancode will require less code, since the translation into a symol will be done from the kernel code and not the scancode).
```
DeanoBurrito, line 105: scancode -> scancodes

Add_Keyboard_Support/DriverImplementation.md
```diff
@@ -0,0 +1,217 @@
...
+        current_state = NORMAL_STATE;
+    }
+}
+```
+
+The code snippet above just show how the states are updated, it doesn't care about the rest. Of course the state machine we have implemente is a very simple one, but if we are going for example to add support for the Print Screen and Pause key, it will probably have more states. The buffer update code can be reused from the previous paragraph. 
+
+### Handling the "Special" keys
+
+When talking about special keys we are referring to those keys that are considered like modifiers keys, when pressed they can alter the output from other keys (we are not talking about the scancode, but for the translated key), or they are simply used in combination with other keys, we are mostly referring to the _Ctrl_, _Alt_, _Shift_, and to some extent also the caps lock (even if this little bit easier to handle).
```
DeanoBurrito, line 139: I would also add the gui/super key to those modifier keys :)

Add_Keyboard_Support/DriverImplementation.md
```diff
@@ -0,0 +1,217 @@
...
+
+When these keys are being kept presed the driver need to know this informantion, because it has to behave accordingly: 
+
+* If the `shift` key is pressed the translation to ascii mechanism needs to know it because it has to return the Capital Symbol associated with that key (for Capital Symbol we mean or the Letter or the symbol above some keys) 
+* If `ctrl` and `alt` are pressed the driver needs to know it because it can trigger either a key combination or some of the "Alt"ernative symbols on some special keyboard keys.
+* If the caps lock key is pressed (not kept pressed) we need the translation function to return only the Capital version of the keys associated with alphabetical letters.
+* The `fn` button is similar to `ctrl` and `alt` but it usually will just trigger an alternate "behaviour" for the keys (** TODO: check if changes the scancode generated by the key**)
```
DeanoBurrito, line 146: fn key is handled by firmware (i.e. the bios). It's not normally visible to the OS. If you do fn+ f12 for example, the os might see it as "volume up" or some other key, not "fn + f12".

Add_Keyboard_Support/InterruptHandling.md
```diff
@@ -0,0 +1,109 @@
...
+|---------|-------------|-----------------------------------------------------------------|
+|  0x60   | R/W         | Data Port                                                       | 
+|  0x64   | R/W         | On read is the Status Register on write is the command register | 
+
+* There are three different sets of scancodes, is a good idea to check with the keyboard what scancode is being used. 
+* Usually the PS/2 controller (the one that the OS usually talks to using ports 0x60 and 0x64) convert set2 scancodes into set1 (for legacy reasons)
+* To check if the translation is enabled, the command 0x20 must be sent on port 0x64, and then read the byte on 0x60. If the 6th bit is set than the translation is enabled. 
+* If we want to disable the translation we need: 
+   - Read current controller configuration byte, with the 0x20 command (the byte will be sent on port 0x60)
+   - Clear the 6th on the current controller configuration byte
```
DeanoBurrito, line 41: 6th -> 6th bit?

Add_Keyboard_Support/InterruptHandling.md
```diff
@@ -0,0 +1,109 @@
...
+   - Send the command 0x60 (it says that next byte on data port has to be written as the controller configuration byte) on port 0x64, then send the updated configuration byte to port 0x60
+   - For our driver we will keep the translation enabled
+* The only scancode set guaranted to be supported by keyboards is the set2. Keep in mind that most of the time the kernel communicate with the 8042_PS2 Controller, and in this case the scancodes can be translated into set1.
```
DeanoBurrito, line 44: the kernel will always talk to the ps2 controller, so you'll always have set2 to set1 translation available. 
The controller is responsible for all the electrical and sometimes mechanical stuff of the ps2 ports, stuff the kernel cant deal with.

Add_Keyboard_Support/README.md
```diff
@@ -0,0 +1,49 @@
+# Adding keyboard support
+
+Writing an OS is great fun, but it would be better if it can interact with the user right, instead of only printing lot of stuff on the screen.
```
DeanoBurrito, line 3: phrasing: Writing an OS is great fun, but instead of only printing stuff to the screen, it would be better if we can interact with the user right?

Add_Keyboard_Support/README.md
```diff
@@ -0,0 +1,49 @@
...
+Whenever a key is pressed or released the keyboard place a code on a communication port, the code is called *scancode*, it can be composed by one or more bytes. 
+
+Currently there are three different sets of scancodes that are available (1, 2 and 3) usually 1 and 2 are the most widely supported and usually the ps/2 controller starts with the translation mode enabled, that means that whatever scancode set is used it will be translated into set 1.
+
+The scancode that is generated when a key is pressed is called the **make** code, while the scancode generated by a key release is calle the **break** code.
+
+So in order to develop our keyboard driver we need to make a set of steps: 
+
+* First identify the scancode set used by our keyboard, this is important because it is going to influence our mapping
+* Enable the Keyboard IRQ, how to do this depends if we are using the PIC or the IOAPIC, but in both cases we need to set up a handler and unmask the relevan entry into a IRQ table
```
DeanoBurrito, line 31: relevan -> relevant

PR #23, opened 2022-05-15
Adding section of notes on compiling a kernel

DeanoBurrito:


dreamos82:
Usually in the  README section i tend to place an overview of the and an initial explanation of what we are going to cover in it. For this reason i usually put a Bullet List with the topic covered by the single files in the section (like in the Memory Management section or the Keyboard section). 
Can you do the same?

dreamos82:
Mayb e we can also split the README in few more subsections?

Build_Process/GNUMakefiles.md
```diff
@@ -0,0 +1,169 @@
...
+
+The line `-rm -r build/` begins with a minus/hyphen. Normally if a command fails (returns a non-zero exit code), make will abort the sequence of commands and display an error. Beginning a line a hyphen tells make to ignore the error code. Make will still tell you an error occured, but it won't stop the executing the make file. In this case this is what we want.
+
+The last two rules tell make how it should create a ``*.c.o` or `*.S.o` file if it needs them. They have a dependency on a file of the same name, but with a different extension (`*.c` or `*.S`). This means make will fail with an error if the source file does not exist, or if we forget to add it to the `SRCS` variables above. We do a protective mkdir, to ensure that the filepath used for output actually exists.
```
inuyasha82, line 84: there is an extra ` here!

Build_Process/GNUMakefiles.md
```diff
@@ -0,0 +1,169 @@
...
+Make has a few built-in symbols that are used through the above example makefile. These are called automatic variables, and are described in the makefile manual.
+
+That's a lot of text! But here we can see an example of a number of make functions being used. This provides a simple, but very flexible build system for a project, even allowing the tools to be swapped out by editing a few lines.
+
+There are other built in functions and symbols that have useful meanings, however discovering them is left as an exercise to the reader.
+
+## Complex Makefile Example (with recursion!)
+What about bigger projects? Well you aren't limited to a single makefile, one makefile can include another one (essentially copy-pasting it into the current file) using the `include` keyword. 
+For example, to include `extras.mk` (.mk is a common extension for non-primary makefiles) into `Makefile` you would add the line somewhere:
+```makefile
```
inuyasha82, line 95: remember the newline when writing codes (is needed by pandoc to render correctly code blocks)

Build_Process/LinkerScripts.md
```diff
@@ -0,0 +1,50 @@
+# Linker Scripts
+If you've never done any kind of low level programming before, it's unlikely you've had to deal with linker scripts. Your compiler will provide a default one for your cpu and operating system. Most of the time this is exactly what you need, however since we are the operating system, we'll need our own!
+
+A linker script is just a description of the final linked binary. It tells the linker how you want the various bits of code and data from your compiled source files (currently they're unlinked object files) to be laid out in the final executable.
+
+For the rest of this section we'll assume you're targetting x86, and are using elf as your file format. If you're building a UEFI binary you'll need to PE/COFF+ as per the spec, and that's a separate thing.
```
dreamos82, line 6: maybe elf should be higlighted somehow? (italic? bold?)

dreamos82: you'll need to PE/COFF+ 

is just my impression or some word is missing?

Build_Process/Overview.md
```diff
@@ -0,0 +1,232 @@
...
+
+A cross compiler is always required when building your os for a different cpu architecture to your host. Building code for an risc-v cpu, while running on an x86 cpu would require a cross compiler for example.
+
+The two main compiler toolchains used are gcc and clang. They differ a lot in philosophy, but are comparable for a lot of the things we care about. GCC is much older and so it established a lot of the conventions used, such as the majority ofs compiler flags, inline assembly and some language extensions. Clang honours most (if not all) of these, and the two seem to be feature equivilent, with the exception of some experimental features.
+
+## Differences Between GCC and Clang
+
+The main difference is that GCC requires a completely separate set of bianries (mean ing a separate build) for each target architecture, while clang is designed in a modular fashion and will swap out the parts for each target architecture as needed. This ultimately means you will need to build (or download) a separate gcc toolchain for each target platform, while clang only needs a single toolchain setup.
```
dreamos82, line 28: bianries -> binaries
ing -> in

Build_Process/Overview.md
```diff
@@ -0,0 +1,232 @@
...
+- `-Wall` and `-Wextra`: These flags need no introduction, they just enable all default warnings, and then extra warnings on top of that. Some people like to use `-Wpedantic` as well, but it can cause some false positives.
+
+### Building C++ Source Files
+This section should be seen as an extension to the section above on compiling C files, compiler flags included.
+
+When compiling C++ for a freestanding environment, there are a few extra flags that are required:
+- `-fno-rtti`: Tells the compiler not to generate **R**un**t**ime **t**ype **i**nformation. This requires runtime support from the compiler libaries, and the os. Neither of which we have in a freestanding environment.
+- `-fno-exceptions`: Requires the compiler libraries to work, again which we dont have. Means you can't use C++ exceptions in your code. Some standard functions (like the `delete` operator) will still required you to declare them `noexcept` so the correct symbols are generated.
```
dreamos82, line 98: will still required..

something missing?

Build_Process/Overview.md
```diff
@@ -0,0 +1,232 @@
...
+
+And a few flags that are not required, but can be nice to have:
+- `-fno-unwind-tables` and `-fno-asynchronous-unwind-tables`: tells the compiler not to generate unwind tables. These are mainly used by exceptions and runtime type info (rtti - dynamic_cast and friends). Disabling them just cleans up the resulting binary, and reduces its file size.
+
+## Linking Object Files Together
+The GCC Linker (ld) and the compatable clang linker (lld.ld) can accept linker scripts.
+These describe the layout of the final executable to the linker: what things go where, with what alignment and permissions.
+Ultimately this file is what's loaded by the bootloader, so these details are be important. More so than they would be in a regular program.
```
dreamos82, line 106: are be important?

Build_Process/BootProtocols.md
```diff
@@ -0,0 +1,85 @@
...
+```
+
+### Fancy Features
+Stivale 2 also provides some more advanced features:
+- It can enable 5 level paging, if requested.
+- It boots up AP (all other) cores in the system, and provides an easy interface to run code on them.
+- It supports KASLR, loading your kernel at a random offset each time.
+- It can also provide things like EDID blobs, address of the PXE server (if booted this way), and a device tree blob on some platforms.
+
+The limine bootloader not only supports x86, but also has tentative ARM (uefi required) support. There is also a stivale2-compatable bootloader called [sabaton](https://github.com/FlorenceOS/Sabaton), providing broader support for ARM platforms.
```
dreamos82, line 82: compatable?

Build_Process/GNUMakefiles.md
```diff
@@ -0,0 +1,160 @@
...
+
+Next up we have flags for the c compiler (`CC_FLAGS`) and the linker (`LD_FLAGS`). If we wanted flags for the assembler, we could a variable here for those too. After the flags we have our first example of where make can be really useful. 
+
+The linker wants a list of compiled object files, from the c compiler or assembler, not a list of the source files they came from. We already maintain a list of source files as inputs, but we dont have a list of the produced object files that the linker needs to know what to link in the final binary. We could create a second list, and keep that up to date, but that's more things to keep track off. More room for error as well.
+Make has built in search and replace functionality, in the form of the `patsubst` (pattern substitution) function. `patsubst` uses the wildcard (`%`) symbol to indicate the section of text we want to keep. Anything specified outside of the wildcard is used for pattern matching. It takes the following arguments:
+- Pattern used to select items from the input variable.
```
dreamos82, line 63: newline here too

Build_Process/GeneratingIso.md
```diff
@@ -0,0 +1,80 @@
...
+## Limine (Stivale 2, multiboot 2)
+Limine requires us to build the iso itself, and then provides a tool to install itself onto the iso. 
+
+To get started we'll want to create a working directory to use as the root of our iso. In this case we'll use `disk/`. Next we'll need to clone the latest binary branch of limine (using `git clone https://github.com/limine-bootloader/limine.git --branch=v3.0-branch-binary --depth=1`) and copy 'limine.sys', 'limine-cd.bin', and 'limine-cd-efi.bin' into `disk/limine/`, creating that directory if it dosn't exist.
+
+Now we can copy our limine.cfg and kernel into your working directory. Limine.cfg needs to be in one of a few locations in order to be found, the best place is the root directory. 
+
+Now we're ready to run the following xorisso command:
+
+```sh
```
dreamos82, line 52: this command is referred to limine case, why it is not added for grub too? 
(or btw since this is kind of common part, it can be placed into a pargraph of it's own otherwise someone using grub can miss it)

Build_Process/GeneratingIso.md
```diff
@@ -0,0 +1,80 @@
...
+xorisso -as mkisofs -b limine-cd.bin -no-emul-boot \
+    -boot-load-size 4 -boot-info-table --efi-boot \
+    limine-cd-efi.bin -efi-boot-part --efi-boot-image \
+    --protective-msdos-label disk -o my_iso.iso
+```
+
+That's a lot of flags! Because of how flexible the iso format is, we need to be quite specific when building one. If you're curious about how crazy things can get, take a look at the flags grub-mkrescue generates behind the scenes.
+
+Now we have an iso will our kernel, config and bootloader files on it, we just need to install limine into the boot partitions, we can do this using `limine-deploy` like so:
```
dreamos82, line 61: will?

Build_Process/GNUMakefiles.md
```diff
@@ -0,0 +1,160 @@
...
+
+The `all` and `clean` targets work as you'd expect, building the final output or cleaning the build files. It's worth noting the '@' symbol in front of echo. When at the start of a line, it tells make not to echo the rest of the line to the shell. In this case we're using echo because we want to output text without the shell trying to run it. Therefore we tell make not to output the echo line itself, since echo will already write the following text to the shell.
+
+The line `-rm -r build/` begins with a minus/hyphen. Normally if a command fails (returns a non-zero exit code), make will abort the sequence of commands and display an error. Beginning a line a hyphen tells make to ignore the error code. Make will still tell you an error occured, but it won't stop the executing the make file. In this case this is what we want.
```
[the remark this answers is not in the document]
DeanoBurrito: That's a fair point. I don't want to encourage people to use `-f` in scripts unless necessary though.

Build_Process/GeneratingIso.md
dreamos82: no you are right. i think in the past i used xorriso to make an iso, but since you mentioned already grub-mkrescue i don´t think there is need for it

Build_Process/BootProtocols.md
```diff
@@ -0,0 +1,85 @@
...
+
+The major difference between multiboot 1 and 2 is how data is communicated between the bootloader and kernel. In multiboot 2 a series of tags (it's a linked list of structs), each one with a pointer to the next tag in the chain.
+
+### Creating a Boot Shim
+The major caveat of multiboot when first getting started is that it drops you into 32-bit protected mode, meaning that you must set up long mode yourself. This also means you'll need to create a set of page tables to map the kernel into the higher half, since in pmode it'll be running with paging disabled, and therefore no translation.
+
+Most implementations will use an assembly stub, linked at a lower address so it can be placed in physical memory properly. While the main kernel code is linked against the standard -2GB address (0xffff'ffff'8000'0000 and above). 
+
+Entering long mode is fairly easy, it requires setting 3 flags:
```
dreamos82, line 38: newline here!

Build_Process/BootProtocols.md
```diff
@@ -0,0 +1,85 @@
...
+To get the next tag in the chain, it's as simple as:
+
+```
+stivale2_tag* next_tag = (stivale2_tag*)current_tag->next;
+if (next_tag == NULL)
+    //we reached the end of the list.
+```
+
+### Fancy Features
+Stivale 2 also provides some more advanced features:
```
dreamos82, line 76: neweline here

Build_Process/LinkerScripts.md
```diff
@@ -0,0 +1,193 @@
...
+
+| Name       | Value | Description |
+|------------|-------|-------------|
+| PT_NULL    | 0     | Just a placeholder, a null value. Does nothing. |
+| PT_LOAD    | 1     | Means this program header describes that should be loaded, in orer for the program to run. The flags specify read/write/execute permissions. |
+| PT_DYNAMIC | 2     | Advanced use, for dynamically linking functions into a program or relocating a program when it is loaded. |
+| PT_INTERP  | 3     | A program can request a specific progam loader here, often this is just the default for the operating system. |
+| PT_NOTE    | 4     | Contains non-useful data, often the linker name and version. |
+
+Modern linkers are quite clever, and will often deduce which program headers are needed, but it never hurts to specify these yourself. For a freestanding kernel you will need at least threee program headers, all of type PT_LOAD:
```
dreamos82, line 51: newline here.

Build_Process/Overview.md
```diff
@@ -0,0 +1,259 @@
...
+## Differences Between GCC and Clang
+
+The main difference is that GCC requires a completely separate set of bianries (mean ing a separate build) for each target architecture, while clang is designed in a modular fashion and will swap out the parts for each target architecture as needed. This ultimately means you will need to build (or download) a separate gcc toolchain for each target platform, while clang only needs a single toolchain setup.
+
+However, each GCC toolchain will use the platform-specific headers by default, where as clang seems to have insane defaults in this area. You'll generally always want to have the platform-specific headers GCC supplies regardless of which toolchain you build with.
+
+Compiling GCC from source doesn't take too long on a modern CPU (~10 minutes for a complete build on a 7th gen intel mobile cpu, 4 cores), however there are also prebuilt binaries online from places like [bootlin](https://toolchains.bootlin.com/).
+
+## Setting up a build environment
+Setting up a proper build environment can be broken down into a few steps:
```
dreamos82, line 35: newline here

Build_Process/Overview.md
```diff
@@ -0,0 +1,259 @@
...
+```
+
+If all goes well, there should be no errors. At this point you can try running your executable, which will likely result in a segfault if its for your native platform, or it wont run if you've compiled it for another platform.
+You can optionally use `readelf` or `objdump` to inspect the compiled elf.
+
+Regarding the flags used above, `-ffreestanding` tells the compiler that this code is freestanding and should not reference outside code. `-nostdlib` tells the linker a similar thing, and tells it not to link against any of the standard libraries. The only code in the final executable now is yours.
+
+Now there are still several things to be aware of: for example the compiler will make the assumption that all of the cpu's features are available. On x86_64 it'll assume that the FPU and sse(2) are available. This is true in userspace, but not so for the kernel, as we have to set them up before they work!
+
+Telling the compiler to not use these features can be done by passing some extra flags:
```
dreamos82, line 81: newline here

Build_Process/Overview.md
```diff
@@ -0,0 +1,259 @@
...
+
+There are also a few other compiler flags that are useful, but not necessary:
+- `-fno-stack-protector`: Disables stack protector checks, which use the compiler library to check for stack smashing attacks. Since we're not including the standard libaries, we cant use this unless we implement the functions ourselves. Not really worth it.
+- `-fno-omit-frame-pointer`: Sometimes the compiler will skip creating a new stack frame for optimization reasons. This will mess with stack traces, and only increases the memory usage by a few bytes here and there. Well worth having.
+- `-Wall` and `-Wextra`: These flags need no introduction, they just enable all default warnings, and then extra warnings on top of that. Some people like to use `-Wpedantic` as well, but it can cause some false positives.
+
+### Building C++ Source Files
+This section should be seen as an extension to the section above on compiling C files, compiler flags included.
+
+When compiling C++ for a freestanding environment, there are a few extra flags that are required:
```
dreamos82, line 96: newline here

Build_Process/Overview.md
```diff
@@ -0,0 +1,259 @@
...
+
+To use a linker script you add `-T script_name_here.ld` to the linker command.
+
+Outside of linker scripts, the linking process goes as you'd expect:
+
+```sh
+$(LD) $(OBJS) -o output_filename_here.elf -nostdlib -static -pie --no-dynamic-linker
+```
+
+For an explanation of the above linker flags used:
```
dreamos82, line 118: newline here

Build_Process/Overview.md
```diff
@@ -0,0 +1,259 @@
...
+- `-Wall` and `-Wextra`: These flags need no introduction, they just enable all default warnings, and then extra warnings on top of that. Some people like to use `-Wpedantic` as well, but it can cause some false positives.
+
+### Building C++ Source Files
+This section should be seen as an extension to the section above on compiling C files, compiler flags included.
+
+When compiling C++ for a freestanding environment, there are a few extra flags that are required:
+- `-fno-rtti`: Tells the compiler not to generate **R**un**t**ime **t**ype **i**nformation. This requires runtime support from the compiler libaries, and the os. Neither of which we have in a freestanding environment.
+- `-fno-exceptions`: Requires the compiler libraries to work, again which we dont have. Means you can't use C++ exceptions in your code. Some standard functions (like the `delete` operator) will still required you to declare them `noexcept` so the correct symbols are generated.
+
+And a few flags that are not required, but can be nice to have:
```
dreamos82, line 100: newline here

Build_Process/Overview.md
```diff
@@ -0,0 +1,259 @@
...
+
+```makefile
+# runs our kernel
+run:
+    qemu-system-x86_64 -cdrom $(ISO_FILENAME)
+run-with-kvm:
+    qemu-system-x86_64 -cdrom $(ISO_FILENAME) --enable-kvm
+```
+
+There are a few other qemu flags you might want to be aware of:
```
dreamos82, line 156: newline here

dreamos82 (commented):
more newlines....

dreamos82 (approved):

Build_Process/BootProtocols.md
```diff
@@ -0,0 +1,280 @@
...
+static stivale2_header stivale2_hdr = 
+{
+    .entry_point = 0,
+    .stack = (uintptr_t)init_stack + 0x2000,
+    .flags = (1 << 1) | (1 << 2) | (1 << 3) | (1 << 4),
+    .tags = (uintptr_t)&framebuffer_tag
+};
+```
+
+If you're not familiar with the `__attribute__(())` syntax, it's a compiler extensions (both clang and GCC support it) that allows us to do certain things our language wouldn't normally allow. This attribute specified that this variable should go into the `.stivale2hdr` section, as is required by the stivale2 spec.
```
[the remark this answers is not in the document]
DeanoBurrito: I think a section on attributes is a great idea! Maybe not as part of this PR, but definitely worth covering.

Build_Process/GNUMakefiles.md
```diff
@@ -0,0 +1,162 @@
...
+build/%.S.o: %.S
+    @echo "Assembling file: $<"
+    @mkdir -p $(@D)
+    $(AS) $< -c -o $@
+```
+
+Okay! So there's a lot going on there. This is just how I like to organise my makefiles, and by no means a definitive guide.
+Since we may be using a cross compiler or changing compilers (it's a good idea to test with both gcc and clang) we've declared some variables representing the various programs we'll call when compiling. `CXX` is not used here, but if you're using c++ it's the common name for the compiler.
+
+Following that we have our inputs, `C_SRCS` is a list of our source files. Anytime we want to compile a new file, we'll add it here. The same goes for `ASM_SRCS`. Why do we have two lists of sources? Because they're going to be processed by different tools (c files -> c compiler, assembly files -> assembly compiler/assembler). `TARGET` is the output location and name of the file we're going to compile.
```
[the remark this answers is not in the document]
DeanoBurrito: Sure that makes sense, it's ultimately a matter of preference so it'd be nice to provide some options.

Build_Process/LinkerScripts.md
```diff
@@ -0,0 +1,196 @@
...
+    rodata PT_LOAD FLAGS((1 << 2));
+    data   PT_LOAD FLAGS((1 << 1) | (1 << 2));
+}
+```
+
+## Sections
+While program headers are a fairly course mechanism for telling the program loader what it needs to do in order to get our program running, sections allow us a lot of control over how the code and data within those areas is arranged.
+
+### The '.' Operator
+When working with sections, we'll want to control where sections are placed in memory. We can use absolute addresses, however this means we'll need to update the linker script manually everytime things change in size. Enter the dot operator (`.`), it represents the current VMA (remember this is the runtime address).
```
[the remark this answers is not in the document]
DeanoBurrito: haha ah it's an idiom! It comes from stage plays, when a character enters the scene. The phrase "Enter x" is usually used as an introduction to something big. 
I suppose for a more technical document there's better ways to word it. I can have a think about it.

Build_Process/LinkerScripts.md
```diff
@@ -0,0 +1,196 @@
...
+
+```
+.text :
+{
+    *(.text*)
+    /* ... more input sections ... */
+} :text
+```
+
+First we give this output section a name, in this case it's `.text`. To the right of the colon is where we would place any extra attributes for this section or override the LMA. By default the LMA will be the same as the VMA, but if you need you can override it here, with something like `.text : AT(0x1234)`. The text section would now have the VMA of 0 unless we've overriden it like before, but LMA of 0x1234. Unless you know you need to change the LMA, let this part empty like the example.
```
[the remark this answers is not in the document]
DeanoBurrito: `AT()` is used for overriding the LMA, correct :)
Linker scripts are focused around changing the VMA, since that's what most programs will be dealing with, so changing the vma is simple: you use the dot operator!

```
. = 0x1234;
```
This would set the current VMA, and the vma will also be incremented by the size of any sections. For example if you added a section after that line above that was 0x1000 bytes, the new VMA would be 0x2234. If you used ALIGN(4K) on the section that followed, it would start at VMA 0x3000.

I can make this clearer in the text.

Build_Process/Overview.md
```diff
@@ -0,0 +1,264 @@
...
+- `-S` pauses the emulator before actually running any code. Useful for attaching a debugger early on.
+- `-s` creates a gdb server on port 1234. Inside of gdb you can attach to this server and debug your kernel/bootloader using `target remote :1234`.
+
+## Building and Using Debugging Symbols
+You'll never know when you need to debug your kernel, especially when running in a virtualized environment. Having debug symbols included in your kernel will increase the file size, but can be useful. If you want to remove them from an already compiled kernel the `strip` program can be used to strip excess info from a file.
+
+Including debug info in the kernel is the same as any other program, simply compile with the `-g` flag. 
+
+There are different versions of DRAWF (the debugging format used by elf files), and by default the compiler will use the most recent one for your target platform. However this can be overriden and the compiler can be forced to use a different debug format (if needed). Sometimes there can be issues if your debugger is from a different vendor to your compiler, or is much older.
```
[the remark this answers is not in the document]
DeanoBurrito: yes it's DWARF. I'm astounded I got that one wrong hahaha

Build_Process/Overview.md
```diff
@@ -0,0 +1,264 @@
...
+
+Getting access to these debug symbols is dependent on the boot protocol used:
+
+### Multiboot 2
+Multiboot 2 provides the Elf-Symbols (section 3.6.7 of the spec) tag to the kernel which provides the elf section headers and the location of the string table. Using these is described below in the stivale section.
+
+### Stivale 2
+Stivale2 uses a similar and slightly more complex (but more powerful) mechanism of providing the location of the entire kernel file in memory. This means you're not limited to just using elf files, and can access debug symbols from a kernel in any format. This is because you have the file base address and length and have to do the parsing yourself.
+
+Parsing elf files is beyond the scone of this section, but it's very straight forward. The elf header contains 2 fields of interest: `e_shoff` (**s**ection **h**eader **off**set) and `e_shstrndx` (**s**ection **h**eader string index).
```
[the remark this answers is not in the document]
DeanoBurrito: Did I mention I'm a chef for my day job? :laughing: Clearly that's where my mind was at the time of writing.

Yep sounds good!

Build_Process/LinkerScripts.md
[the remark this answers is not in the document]
dreamos82: ah ok! I didn't know! Ok so it is fine :) 
I see now what it means (i think there is something similar in italian, but i didn't associated it with the idiom) maybe just use italic to highlight or quotes to higlight the idiomatic expression :)

Build_Process/LinkerScripts.md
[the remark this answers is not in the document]
dreamos82: yeah please! :)

Build_Process/BootProtocols.md
```diff
@@ -0,0 +1,280 @@
...
+
+The major difference between multiboot 1 and 2 is how data is communicated between the bootloader and kernel. In multiboot 2 a series of tags (it's a linked list of structs), each one with a pointer to the next tag in the chain.
+
+### Creating a Boot Shim
+The major caveat of multiboot when first getting started is that it drops you into 32-bit protected mode, meaning that you must set up long mode yourself. This also means you'll need to create a set of page tables to map the kernel into the higher half, since in pmode it'll be running with paging disabled, and therefore no translation.
+
+Most implementations will use an assembly stub, linked at a lower address so it can be placed in physical memory properly. While the main kernel code is linked against the standard -2GB address (0xffff'ffff'8000'0000 and above).
```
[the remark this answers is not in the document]
DeanoBurrito: It's explained below, but I will move it up here.

Build_Process/Overview.md
```diff
@@ -0,0 +1,264 @@
...
+The elf section headers share a common format describing their contents. They can be thought of as an array of `Elf64_Shdr` structs.
```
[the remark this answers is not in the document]
dreamos82: yeah i get it, but maybe just saying what does the struct contains at a high level couldn't be little bit beneficial? 

There is no need for the whole struct (especially if it is big) but, something like is in the Elf_Shim structure, this structure contains information about ... 

What do you think?

dreamos82:
another small comment, i was looking at the environment setup, and there was this section: 

Some symbols often used in our makefile:

$@ is the name of the target being generated
$< the first prerequisite
@D is the target directory


where was explaining the usage of some symbols, i was wondering if it can be wroth adding a small subsection explaining the most common used symbols in the makefile for our purpose (since they are also used in the exmaple makefile)

Build_Process/BootProtocols.md
```diff
@@ -283,3 +294,56 @@ void kernel_start(stivale2_struct* stivale2_data);
...
+    {
+        if (tag->type == 0 && size == 8)
+            return NULL; //we've reached the terminating tag
+        
+        if (tag->type == type)
+            return tag;
+        
+        tag = (multiboot_tag*)((uintptr_t)tag + tag->size);
+    }
+}
```
inuyasha82, line 321: shouldn´t we return null in case we don't find our tag?

Build_Process/BootProtocols.md
```diff
@@ -0,0 +1,349 @@
...
+Since both multiboot 2 and stivale 2 return their info in linked lists, a brief example of how to traverse these lists is given below. These functions provide a nice abstraction to search the list for a specific tag, rather than manually searching each time.
+
+### Multiboot 2
+Multiboot 2 gives us a pointer to the multiboot info struct, which contains 2x 32-bit fields. These can be safely ignored, as the list is null-terminated (a tag with a type 0, and size of 8). The first tag is at 8 bytes after the start of the mbi. All the structures and defines used here are available in the header provided by the multiboot specification (check the bottom section, in the example kernel).
+
+```c
+//placed in ebx when the kernel is booted
+multiboot_info* mbi;
+
+void* multboot2_find_tag(uint32_t type)
```
dreamos82, line 308: multboot?

Build_Process/BootProtocols.md
```diff
@@ -0,0 +1,349 @@
...
+        if (tag->type == type)
+            return tag;
+        
+        tag = (multiboot_tag*)((uintptr_t)tag + tag->size);
+    }
+}
+```
+
+The last line of the loop is a little messy, in fact most compilers will let you omit the final cast to the `void*`, and will convert the integer expression directory to a pointer for you (gcc and clang both allow this).
+
```
dreamos82, line 325: In the original multiboot file:
**Tags follow one another padded when necessary** in order for each tag to start at 8-bytes aligned address. So when iterating through tags we need always to align the address to 8byte, this explain the for loop in the multiboot2 example:

```
  for (tag = (struct multiboot_tag *) (addr + 8);
       tag->type != MULTIBOOT_TAG_TYPE_END;
       tag = (struct multiboot_tag *) ((multiboot_uint8_t *) tag 
                                       + ((tag->size + 7) & ~7)))
```
And the text says: 
To get to the next tag, we always need to add the size of the current tag (that is in tag->size) and align it to 8bytes. this is done by adding 7 to the current tag size, and then AND the result with negation of 7.

Tags are terminated by a tag of type ‘0’ and size ‘8’.

One thing, i'm totally ok with the while loop, but please mention the aligned  increment part (i took it for myself but it took time for me to figure out what was the meaning of this for loop increment! :D And why that + 7 & ~7

But what i was wondering is that it says that the header are padded when necessary, So in theory the while could not always end at the beginning of the next item if it doesn't take it into account? (maybe you should keep the aligned increment in the while too?)

Build_Process/BootProtocols.md
```diff
@@ -0,0 +1,349 @@
...
+### Stivale 2
+Stivale 2 gives us a pointer to a header at the start of the list, and then each item (including this header) contains a `next` pointer to the next item, and an `id` item with a unique 64-bit identifier for that tag. All the structures and defines are available in the standard `stivale2.h`. We'll know we've reached the end of the list when the `next` pointer is `NULL`.
```
dreamos82, line 327: maytbe is worth adding a similar sentence on the multiboot paragraph too? mentioning that there is a header available ready to use, and all the types and multiboot struct are declared there?

Build_Process/BootProtocols.md
```diff
@@ -0,0 +1,349 @@
...
+        if (tag->id == id)
+            return tag;
+        tag = tag->next;
+    }
+
+    return NULL;
+}
+```
+
+The above function can be used with the defines in `stivale2.h`, which follow the format `STIVALE2_STRUCT_TAG_xyz_ID`, where `xyz` represents the feature that is described in the tag. For example, the framebuffer would be `STIVALE2_STRUCT_TAG_FRAMEBUFFER_ID` and the memory map is `STIVALE2_STRUCT_TAG_MEMMAP_ID`. It's a little verbose, but easy to search for.
```
dreamos82, line 349: maytbe is worth adding a similar sentence on the multiboot paragraph too? mentioning that there is a header available ready to use, and all the types and multiboot struct are declared there?

PR #26, opened 2022-06-02
Minor changes timer nasm

dreamos82:


Nasm.md
```diff
@@ -252,3 +252,68 @@ je .value3_case
 ```
 
 ## Data structures
+
+As you can imagine assembly doesn't have intrinsic concept of data structures, but luckily we have the support of the preprocessor where that let us declare struct like blocks.
```
DeanoBurrito, line 256: I would say this is the other way around: Every language supports accessing data as a raw array of bytes, C provides an abstraction over this in the form of structs. NASM also happens to provide us with an abstraction over raw bytes, that is similar to how C does it.

Nasm.md
```diff
@@ -252,3 +252,68 @@ je .value3_case
...
+```
+
+How nasm render a struct is basically declaring a list of offset labels in this way  we can use them to access the field starting from the struct memory location (*Authors note: yeah it is a trick...*)
+To create a struct in nasm we use the `struc` and `endstruc` keywords, and the fields are defined between these them. 
+The example above can be rendered in the following way:
+
+```asm
+struc task
+    id:         resd    1
+    name:       resd 8
```
DeanoBurrito, line 276: This shold be `resb` I think :P

Nasm.md
```diff
@@ -252,3 +252,68 @@ je .value3_case
...
+endstruc
+```
+
+what this code is doing is creating three symbols: id as 0 representing the offset from the beginning of a task structure and name as 4 (still the offset) and the task symbol that is 0 too. This notation has a drawback, it defines the labels as global constants, so you can't have another struct or label declared with same name, to solve this problem you can use the following notation:
```
DeanoBurrito, line 280: what -> What

Nasm.md
```diff
@@ -252,3 +252,68 @@ je .value3_case
...
+what this code is doing is creating three symbols: id as 0 representing the offset from the beginning of a task structure and name as 4 (still the offset) and the task symbol that is 0 too. This notation has a drawback, it defines the labels as global constants, so you can't have another struct or label declared with same name, to solve this problem you can use the following notation: 
+
+```asm
+struc task
+    .id:    resd    1
+    .name:  resb    8
+endstruc
+```
+
+In this case the ofssets label created are `task.id`, `task.name`. So to put it in simpler terms what does we really have here are some label that represents offset, and to access the fields of a structure what we do is adding these offsets to the memory location containing it.
```
DeanoBurrito, line 289: This whole line can be simplified a bit I think: 
Now we can access the fields inside our struct in a familiar way: `struct_name.field_name`. What's really happening here is the assembler will add the offset of `field_name` to the base address of `struct_name` to give us the real address of this variable.

Nasm.md
```diff
@@ -252,3 +252,68 @@ je .value3_case
...
+
+Now if we have a memory location or register that contains our structure, for example let's say that we have the pointer to our structure stored in the register rax and we want to copy the id field in the register rbx:
+
+```nasm
+mov rbx, dword [(rax + task.id)]
+```
+
+This is how to access a struct, besically we add the label representing an offset to its base address.
+Bbut what if we want to create an instance of it? Well in this case we can use the macros `istruc` and `iend`, and using `at` to access the fields. For example if we want create an instance of task with the values 1 for the id field and "test" for the name field, we can use the following syntax:
```
DeanoBurrito, line 298: Bbut -> But

Nasm.md
```diff
@@ -252,3 +252,68 @@ je .value3_case
...
+
+In this way we have declared a struc for the first of the two examples. But again this doesn't work with the second one, because the labels are different. In that case we have to use the full label name (that means adding the prefix task):
+
+```asm
+istruc task
+    at task.id      dd 1
+    at task.name    db 'hello123'
+iend
+```
+
```
DeanoBurrito, line 315: One other thing that should probably also be mentioned is using structs before c and assembly: there's a lot that can go into that (for example padding, or if you've packed the struct in c, struct alignment). This is more related to C (so we could put it in the C notes), but I'd say its worth leaving a sentence about it here.

Nasm.md
dreamos82: yeah lol i originally wrote resb, than changed the strucuture few times to end back to returning to the first version, but forgot to change it back XD

PR #29, opened 2022-06-22
Scheduling and tasks

dreamos82:


04_Scheduling/02_Scheduler.md
```diff
@@ -0,0 +1,149 @@
...
+### Next stuff to be completed...
+
+Now if the `current_thread` har reached it's allocated time, it's time to pick the next one. But before doing that we need to save make sure that next time `current_thread` will be picked up, it will resume from the exact point it is being interrupted. This is achieved by saving the current execution context. And what is it? Well we already encountered that, in the [interrupt handling](../InterruptHandling.md) chapter, it is the status of the cpu in that exact istant, all the registers value (the instruction pointer, the stack values, the general purpose registers etc). And this gives us a big hint on how we are going to switch between threads: we will avail of our interrupt handler (*Authors note: of course this is not the only way to switch between task, but this is in our opinion one of the easiest to implement).*)
+
+Let's recall quickly what our interrupt handling routine does: 
+
+* It first save the context pushing all registers on the stack
+* Then it calls the interrupt handling function, that serves the interrupt
+* It restore the context from the stack
+* Return (if possible) the control to the kernel
```
DeanoBurrito, line 141: I see why you've put the 'if possible', but I think it'd be more accurate to rewrord this. Since we're returning to whatever the iret frame indicates (could be user code, could be user code, could even be another interrupt handler if we support nested interrupts). Perhaps word it like:

Returns control to the previously interrupted code.

04_Scheduling/03_Tasks_And_Threads.md
```diff
@@ -0,0 +1,132 @@
...
+
+** THIS SECTION IS IN EARLY STAGES For now it will be more a set of bullet lists with things to be expaneded in the future ** 
+
+## Tasks
+
+### What are tasks
+
+In a task usually there are the following information: 
+
+* Virtual memory base
```
DeanoBurrito, line 11: I know it's early days, but I think the bare minimum for a task is just an id, list of threads and virtual memory manager. Everything else is optional, even the name.
For a thread all you really need is an id, its parent task, and the current frame. Of course it could also be useful to keep track of the stack base (so you can free it when the thread is dead).

04_Scheduling/03_Tasks_And_Threads.md
```diff
@@ -0,0 +1,132 @@
...
+
+### Prerequisites 
+
+Even if technically not necessary, is better to have implemented a memory allocation mechanism, ideally both physical and virtual memory manager should be implemented, if we want to have full tasks/thread separation and protection. But in case we don't want to have memory separation and protection in place a physical memory manager can be enough.
+
+Technically we could also decide to implement a "kind of" multitasking without any memory allocator, but in this case we must use an array with a fixed size. So the Operating system will be limited to a certain number of concurrent task, and this memory will be always be unavailable to the os even when the tasks are dead.
+
+### Creating a thread
+
+When creating a thread we must first of all create it's execution frame (yeah even if not executed yet, it will need one).
```
DeanoBurrito, line 37: As a reader, I would ask why?
It can be useful to explain (here, or above when first talking about the execution context) what the context is. How do you start running *new* code on the cpu? The osdev makes an excellent suggestion of making the cpu think it was already running it, by creating a new iret frame, and using that instead of the current one. Thus you return to new code.

04_Scheduling/03_Tasks_And_Threads.md
```diff
@@ -0,0 +1,132 @@
...
+
+### What a task should contain ?
+
+So what are the information that we should store in a task? That depends on the design choices, because if we are implementing a multitask/single thread os it will need more information, but if we are going to implement a multi-task/multi-threaded OS some of the information will be stored in the Thread structure. So let's start with the minimum set of information that we need to save into a task: 
+
+* First of all a name, all tasks have a task name, most of the time is the excuatble name, sometime something different
+* Every task will have their own task id, this value usually is used by the scheduler to pick up the next task to execute or and in all the scenarios where an access to the task is needed (for example interprocess comunication)
+* A reference to it's own virtual addressing space, that is a pointer to a PML4 table (this only if we are going to implement an os where every task has it's own address space)
```
DeanoBurrito, line 49: I would say it's pretty dangerous not to isolate tasks within their own virtual spaces.  Even if it makes stuff like IPC easier, you have no protection from other programs. What happens if 2 programs want to be loaded at the same address? What if neither or them can be relocated? Boom.

04_Scheduling/03_Tasks_And_Threads.md
```diff
@@ -0,0 +1,132 @@
...
+### Context Switching
+
+* Usually context switch is done during an IRQ (when it depends on design decision. 
+
+One way to achieve context switching, is to clone the current stack frame for the IRQ being served and update some of it's values:
+
+* rip - this will contain the pointer to the thread function
+* rflgas - will containg 0x202
+* rsp - is the stack pointer, since the pointer is growing backward 
+* // to add other registers to initialize
```
DeanoBurrito, line 60: just missing `ss` and `cs` :) These can be the selectors (not segment offsets, they have the RPL and table fields in the lower 3 bits), described in the GDT chapter.

04_Scheduling/03_Tasks_And_Threads.md
```diff
@@ -0,0 +1,132 @@
...
+```c
+void thread_execution_wrapper( void (*function)(void *), void *arg) {
+    function(arg);
+    _thread_exit();
+}
+```
+
+What should the exit function do? Again this depends on the design choice, and as usual there are multiple paths, it depends if we want to exit the thread as soon as it calls the the exit function, or let the scheduler do that, just updating its status to the one correspinding to a terminated task (let's call it DEAD status). 
+
+* If the choice is to delete the thread as soon as it exit terminate the function what will happen then is that the thread is placed into a DEAD state, then the function will take care of removing the task from the scheduler queue, freeing all resources (stack, execution frame, page tables, etc.) that are allocated to it, and after that it will free the memory allocated to the task itself. Remember that when a task call the exit function is still the one being executed
```
DeanoBurrito, line 87: I see what you're hinting at here, maybe its worth overtly saying that the danger here (as I found out haha) is that if you free the stack of the current thread, and your scheduler does not switch stacks, you will free the stack the cpu is currently using! You must be super careful with this approach, I think it might even be worth suggesting using the second option (what you use in dreamos64).

04_Scheduling/03_Tasks_And_Threads.md
```diff
@@ -0,0 +1,132 @@
...
+
+Now when a function is  called, the cpu put the next instruction on the stack. And here is where we will put the end function.
+
+* Then the thread_exit function should take also care of updating the thread status to DEAD (and by extension if it was the last thread on a task it should update that too)
+* The schedule function when pick a task that has the status set to DEAD knows that it doesn´t have to execute it, and will call the routine to free the thread resources. 
+* If the thread is the last in a task, the task should be ready to be deleted too. So in this case we need to change the status to the task too. 
+ 
+## Switching tasks/thread
+
+So who is responsible of switching tasks? The scheduler, that every time is called it checks if the current task needs to be replaced with a new one, and eventually pick the "best" next task candidate. How to decide who is the next depends on the scheduling algorithm implemented and the design choice of the operating system (there are many scheduling algorithm).
```
DeanoBurrito, line 112: This is a repeat of parts of the scheduling chapter. I'd also say that thread selection should be part of the scheduler, not tasks and threads.

04_Scheduling/01_README.md
```diff
@@ -0,0 +1,16 @@
...
+This is where a scheduler comes in. It's similar to a hardware multiplexer (sometimes called a mux), in that it takes multiple inputs (our code) and allows them to share a single output (our cpu, which can only run one instruction stream). 
+
+The scheduler works by interrupting the currently running program, saving it's state and then loading the state of a new program. This is called a context switch, and if often quite an expensive and slow operation to perform, as we'll see later. As you might have guessed, the scheduler often operates within an interrupt handler, and if written correctly, the running program will never know it has paused at all. If these context switches are performed at the right speed, the user will never notice programs briefly pausing and unpausing, and it will appear to the user as if all programs are running at the same time.
+
+Of course unless you support multiple cores, we know this is far from the truth. It's all just a clever trick.
```
DeanoBurrito, line 8: I know I wrote this, but now I'm thinking that perhaps a good analogy could be how videos work: It's just a sequence of still images played rapidly enough that we perceive it as smooth motion.

04_Scheduling/02_Scheduler.md
```diff
@@ -154,14 +154,64 @@ void schedule() {
...
+    tasks_list[current_executing_task_idx].context = context
+    current_executing_task_idx = (current_executing_task_idx + 1) % MAX_TASKS;
+    while(tasks_list[current_executing_task_idx] == NULL) {        
+        current_executing_task_idx = (current_executing_task_idx + 1) % MAX_TASKS;
+    }
+    return tasks_list[current_executing_task_idx].context;
+}
+```
+
+And now our scheduler is capable of switching tasks. /* should i mention that this should work only if the rsp is changed with the new stack? */
```
DeanoBurrito, line 213: I would say yes, but you did mention above that this expects the reader has implemented an interrupt handler similar to one we implemeneted in the interrupts chapter. I think that's enough of a hint, but if you want to be sure a sentence on it here wouldn't hurt :)

04_Scheduling/02_Scheduler.md
```diff
@@ -154,14 +154,64 @@ void schedule() {
...
 
-In the previous section we have seen how to iterate throught tasks, but that basically just iterating through items of a list, so for now our scheduler is still doing nothing. Once the next task to be executed has been selected selected, the first thing that the scheduler has to do is stop the current task, and save its current status. If you have implemented the interrupt handler following this guide you should already have the context being passed to the it, so now we need to change the signature of our schedule function, adding a new parameter: 
+In the previous section we have seen how to iterate throught tasks, but that basically was just iterating through items of a list, not very useful. 
+
+#### Saving Context
+
+Once the next task to be executed has been selected, the first thing that the scheduler has to do is stop the current task, and save its current status. If you have implemented the interrupt handler following this guide you should already have the context being passed to, what we want is to pass down the context to the schedule function, but first  we need to change its signature of our schedule function, adding a new parameter:
```
DeanoBurrito, line 163: I would say the scheduler doesn't *stop* the current task, as it's already been paused when the interrupt fired. The task is not running in the background while the interrupt runs, the interrupt firing has already stopped the task.

04_Scheduling/02_Scheduler.md
```diff
@@ -154,14 +154,64 @@ void schedule() {
...
+
+that's it, this is how how we save the status of the current executing task. 
+
+#### Restoring context
+
+After the previous task status has been saved and the next task has been selected we are now ready to make the switch. How to do it is pretty straightforward, but let's have a quick recap on how we are handling interrupts: 
+
+1. The first thing is that whenever an interrupt is fired an asm code snippet is called saves the current cpu registers status, on the stack, 
+2. Then once the status is saved, it place the current stack pointer (`rsp`) on the `rdi` register (it is the first c function parameter for x86_64 architecture) and call the interrupt handler. This parameter is our `context` variable (that is passed also to the scheduler)
+3. Then our interrupt handler function (unless we are writing an os entirely in asm, this usually is on our language of choice, in our case C)  does whatever it needs to serve the interrupt  and  the context variable. Returning a variable, according to the ABI calling convention is corresponding to placing the returning value to the `rax` register
```
DeanoBurrito, line 190: although funny, I think the 'os entirely in asm' comment is not needed.

04_Scheduling/02_Scheduler.md
```diff
@@ -154,14 +154,64 @@ void schedule() {
...
+4. Now we are back in the asm code where `rax` value is placed on the `rsp` register, and then the context restored...
```
DeanoBurrito, line 191: why `rax`? It's obvious to us that it's the sysv x86_64 calling convention, but might be worth a quick mention in brackets.

04_Scheduling/03_Tasks_And_Threads.md
```diff
@@ -0,0 +1,132 @@
...
+
+Technically we could also decide to implement a "kind of" multitasking without any memory allocator, but in this case we must use an array with a fixed size. So the Operating system will be limited to a certain number of concurrent task, and this memory will be always be unavailable to the os even when the tasks are dead.
+
+### Creating a thread
+
+When creating a thread we must first of all create it's execution frame (yeah even if not executed yet, it will need one).
+
+### Execution frame
+
+One of the way to implement stack switching is replacing the current iret frame with a the one contained in the next thread. To do that a task need when created to build an ad-hoc frame that will not cause any crash. What it should contain? Well this is basically a copy of the stack frame during the handling of an interrupt routine.
```
DeanoBurrito, line 41: I believe the intel and amd manuals call this the 'iret frame'. While I think using our own ideas and terms is nice in some places, what you're talking about here is well defined and giving it another name might be confusing for new readers.

04_Scheduling/03_Tasks_And_Threads.md
```diff
@@ -0,0 +1,132 @@
...
+* Stack pointer
+* Execution frame
+* Pointer to function to run
+* Eventually arguments for the function
+* Depending on how the scheduler is designed, it could contains a pointer to the next thread
+* Task containing the current thread
+
+### Prerequisites 
+
+Even if technically not necessary, is better to have implemented a memory allocation mechanism, ideally both physical and virtual memory manager should be implemented, if we want to have full tasks/thread separation and protection. But in case we don't want to have memory separation and protection in place a physical memory manager can be enough.
```
DeanoBurrito, line 31: While possible to implement scheduling without a vmm (only using pmm), I don't think it's an idea we should encourage. It'll likely lead to bad design decisions (tasks allocating physical memory instead of working with virtual addresses), and also implies that they dont have a heap.
I think it's safe to assume that the reader is reading things in order, and has therefore read the memory/allocators section.

04_Scheduling/03_Tasks_And_Threads.md
```diff
@@ -0,0 +1,132 @@
...
+
+### Execution frame
+
+One of the way to implement stack switching is replacing the current iret frame with a the one contained in the next thread. To do that a task need when created to build an ad-hoc frame that will not cause any crash. What it should contain? Well this is basically a copy of the stack frame during the handling of an interrupt routine. 
+
+### What a task should contain ?
+
+So what are the information that we should store in a task? That depends on the design choices, because if we are implementing a multitask/single thread os it will need more information, but if we are going to implement a multi-task/multi-threaded OS some of the information will be stored in the Thread structure. So let's start with the minimum set of information that we need to save into a task: 
+
+* First of all a name, all tasks have a task name, most of the time is the excuatble name, sometime something different
```
DeanoBurrito, line 47: Task name is nice to have, but definitely not required info to store about a task.

04_Scheduling/03_Tasks_And_Threads.md
```diff
@@ -0,0 +1,132 @@
...
+The rip will contain the function that is associated with the thread, sp contains the newly allocated stack for it. 
+
+### Exiting the thread
+
+After the thread finish its execution, we need a way to make it exit gracefully (otherwise it start to run into garbage most likely...) so there are two possible scenarios:
+
+* The programmer has called a thread_exit function so in this case we are fine
+* The programmer didn't called the function and the thread has terminated finished the execution of the function, at this point if it will not be stopped it will run into garbage. 
+
+To achieve that we need to have an "exit" function to be executed after the function in the thread terminates it's execution. But how to do it? We need to check on the X86_64 Abi calling convention (or what architecture applies to you). As usual there are multiple ways to achieve that, the easiest one is to create a wrapper function that takes two parameters
```
DeanoBurrito, line 71: I would split this into two separate steps:
- first you write the exit function, and supporting code. Show how it wold work with a thread calling it manually.
- Now you can introduce the idea of the wrapper function, and have it automatically exit when a thread's entry function returns.

04_Scheduling/03_Tasks_And_Threads.md
```diff
@@ -62,13 +62,109 @@ typedef struct
 
 Where `NAME_MAX_LEN` is a constant with the maximum length for our process it´s up to us. The pid is a unique number, it can even be a uuid if we want, but the easiest way to achieve this is just using sequential number that starts from zero. To achive this we just need to add a global variable that contains the next available number. Now since we are using size_t on a 64bit architecture, we don't need to worry about overflowing the variable size, technically we can decide to reuse pids that are no longer used by a process, that is totally up to us, but in our simple scenario we will increment a global variable: 
 
-```
+```c
 size_t next_free_pid = 0;
 ```  
 
+
 ### Virtual memory space
```
DeanoBurrito, line 70: This is a title, so each word gets capitalized: Virtual Memory Space

04_Scheduling/03_Tasks_And_Threads.md
```diff
@@ -62,13 +62,109 @@ typedef struct
...
 
-TBD 
+On of the most useful features of having the paging enabled is that we have the concept of Virtual Memory available (for a complete definition refer to the [Paging](,,/02_Memory_Management/Paging.md) and [Virtual Memory](../02_Memory_Management/04_Virtual_Memory_Manager.md) chapter), and from a process point of view it means that we can have each process with it's own addressing space.
```
DeanoBurrito, line 72: `Virtual Memory` is not a title or proper noun so it doesn't get capitalized here (there are other examples of this, I'm just using these two).

04_Scheduling/03_Tasks_And_Threads.md
```diff
@@ -62,13 +62,109 @@ typedef struct
...
+    status_t process_status;
+    cpu_status_t context;
+    uint64_t pdbr;
+} process_t;
+
+```
+
+Adding the pdbr reference is not enough. Let's see why:
+
+* The first problem is that the PDBR contains the pointer to the root page directory of the virtual memory space (remember that an address with pagin enabled is a composition of table indexes), so this means we need to allocate for it.
```
DeanoBurrito, line 94: pagin -> paging

04_Scheduling/03_Tasks_And_Threads.md
```diff
@@ -62,13 +62,109 @@ typedef struct
...
+* A memory address to be accessibile, has to have the present flag set, and various directories/table entries to be properly configured
+* The step above means that when we create a new virtual memory space, we lose any reference to the kernel, so this means that as soon as we switch to the new pdbr, if the kernel is not mapped where it is suppose to be we will cause our kernel to crash, so we need to remap the kernel in the new process memory environment. How to map it is explained in the Paging chapter.
+
+Implementing them is just a variation of what we have seen in the Paging chapter, so we'll leave it as an exercies.
```
DeanoBurrito, line 98: Nice summary! Although this last line says the same thing as the previous line, it can probably be omitted.
There's also a few sneaky issues that can arise from having a separate page table per process later on. For example, if the kernel in one process maps something in the higher half, it will only appear in that process. If the kernel of another process (it's the same kernel, but with a different mapping now - because its a different process) tries to access a pointer in that region it will page fault. There's a bunch of ways to solve this, and it could probably get it's own section. Maybe I can write something on what I've tried so far.

04_Scheduling/03_Tasks_And_Threads.md
```diff
@@ -62,13 +62,109 @@ typedef struct
...
+    status_t process_status;
+    cpu_status_t context;
+    uint64_t pdbr;
+    Heap_Node* heap_root;
+    Heap_Node* cur_heap_position;
+} process_t;
+
+```
+
+You can notie that we have actually added two variabiles, this is because of our memory allocation algorithm.
```
DeanoBurrito, line 119: notie :laughing:

04_Scheduling/03_Tasks_And_Threads.md
```diff
@@ -62,13 +62,109 @@ typedef struct
...
+    int *a = malloc(sizeof(int));
+```
+
+* Proess 2 make the following call:
+
+```c
+    custom_struct *cs = malloc(sizeof(custom_struct))
+```
+
+And in both cases the address returned is 0x10000. Apparently they are the same. But the pdb address in process 1 is different from pdbr address in process 2, and this means that we will have different page directories/and tables used. Since they are different we are physically referring to two different items, and they are not related so the will contain two different physical addresses (yes we can in theory having them pointing to the same physical address, and in some case we will also do it, for example when mapping the kernel, but is our physical memory manager duty to make sure that we don't allocate the same physical address twice).
```
DeanoBurrito, line 165: Nice example!
We should add that this is probably not a good idea of how to do it in practice. For the kernel you will want a single heap, unless you can be certain of which process will be running when some memory is accessed. Quite often in the kernel you will want to access global state, meaning it must be visible to all processes, and therefore can't be stored on a per-process heap. There are of course exceptions to this, but I haven't come across enough for a per-process heap to be worth it.
For a user program they will provided their own heap (either custom, or the one part of the standard library they linked against). This operates in userspace so there isn't the massive overhead of a system call with every alloc/free call.

04_Scheduling/04_Locks.md
```diff
@@ -0,0 +1,205 @@
...
+
+```c 
+#define SHARED_RESOURCE 0xDEADBEEF 
+
+// Other code doing other stuff
+char string_to_send[] = "I am the first string"
+
+int i = 0;
+while(i < strlen(string_to_send) {
+    *((int *) shared_resource) = string_to_send[i++];
```
DeanoBurrito, line 21: It'd probably make more sense for this to be a `char*` I think :P

04_Scheduling/04_Locks.md
```diff
@@ -0,0 +1,205 @@
...
+Imagine that we have a shared resource that can be accessed at a specified address (that resource can be anything: an i/o memory mapped device, a memory mapped hard drive location, a shared variable, or whatever).
+
+Let's assume for example that the shared resource is accessible at the address 0xDEADBEEF, this resource is a serial port connected to an output device. 
+
+Now we have initially a single process that is sending a string to the serial port one character at time: 
+
+```c 
+#define SHARED_RESOURCE 0xDEADBEEF 
+
+// Other code doing other stuff
```
DeanoBurrito, line 16: I think for the sake of brevity we dont need these comments

04_Scheduling/04_Locks.md
```diff
@@ -0,0 +1,205 @@
...
+IWahmi....
+```
+
+While we should expect: 
+
+```
+I am the first string While i am the second
+```
+
+The above scenario happens because both processes (A and B)  have access to the shared resource at the same time, and both of them are writing to it. The junk string above is an example of _race condition_, they are situation in which memory location are accessed concurrently, and at least one of the access is a write. We can imagine that the scenario above can lead to many problems so what we need is a mechanism to protect the shared resource. This is where we introduce the concept of locking.
```
DeanoBurrito, line 60: Nice definition! I'm going to nitpick a little and say that a race condition is a situation where that access has side-effects. Although it's almost always going to be a write, there are some read operations can side effects. A few of the registers in the NVMe protocol come to mind, and so does the risc-v version of the io-apic.

04_Scheduling/04_Locks.md
```diff
@@ -0,0 +1,205 @@
...
+But how can we ensure mutual exclusion then? The basic idea is pretty straight forward, imagine we update both processes to use the acquire and release function while accessing the the shared resource(be aware in order to have a functionig lock mechanism we should have some kind of memory allocation mechanism implemented). This is the first process:
+
+```c
+ //Process A
+#define SHARED_RESOURCE 0xDEADBEEF 
+
+// Other code doing other stuff
+char string_to_send[] = "I am the first string"
+
+spinlock_t *lock = malloc(sizeof(spinlock_t));
```
DeanoBurrito, line 94: Alternatively this could be allocated on the stack:
```
spinlock_t lock;

acquire(&lock);
```
As it is currently, `&lock` will also pass ` spinlock_t**` to acquire.

DeanoBurrito: because you take the address of a pointer, that is*

04_Scheduling/04_Locks.md
```diff
@@ -0,0 +1,205 @@
...
+``` 
+
+and below the second one: 
+
+```c
+#define SHARED_RESOURCE 0xDEADBEEF 
+// Other code doing other stuff
+char string_to_send[] = "While i am the second"
+
+spinlock_t *lock = malloc(sizeof(spinlock_t));
```
DeanoBurrito, line 112: This implementation won't actually protect the shared resource: you have created two *different* locks. So they can both be acquired and allow access, and are functionally independent.

DeanoBurrito: A easy solution is to remove this instance of the lock, and then the code that follows will reference the first instance of the lock.

04_Scheduling/04_Locks.md
```diff
@@ -0,0 +1,205 @@
...
+
+* It doesn't guarantee the order of the processes requiring access to it. So if processes A, B and C require access to the shared resource, they are not guaranteed to be granted the lock in the same order, and in some scenario this could be a problem
+* It also doesn't prevent a _deadlock_ to happen. A _deadlock_ is a scenario where there are two or more processes that are waiting to acquire a shared resource that is hold by the other one. For example if A is holding resource X and needs to acquire a lock on Y, while B is holding a lock on Y and needs to acquire a lock on X, if not handled correctly we will have A and B stuck in the acquire loop forever.
+
+
+Solution to the problems above (and others that can arise) is a research area and out of scope of this guide, but is good to know that there are many algorithms available that are target to solve or mitigate these issues, and could be a good idea to add few of them to our kernel. 
+
```
DeanoBurrito, line 204: Very nice! I would also mention semaphores, and locks that interact with the scheduler. You could have a lock that if it can't acquire, it yields to the scheduler, rather than spinning. This is less responsive, but saves cpu time.
I know you mentioned deadlocks, but there is a special kind of deadlock that can occur in kernels with interrupts. Say a lock is taken (like in the PMM), and then an interrupt occurs. Now maybe the interrupt handler needs to allocate from the PMM, we have a deadlock. The solution to this is to disable interrupts anytime a lock is taken that might be accessed from an interrupt.

04_Scheduling/04_Locks.md
```diff
@@ -0,0 +1,205 @@
...
+* ... etc.
+
+As we can see this simple algorithm has prevented two processes to interfere with each other. So can we assume that now we are safe? Well... no, there is one problem with the above implementation, this doesn't ensure mutual exclusion in a multi processor environment. Imagine we have 2 different processors calling the acquire at the same time (when we have a multi-cpu environment we have several processes/thread that run simultaneously) and they reach the if statement at the same time with the locked variable still to false, they will both reclaim the lock and will start to write to the shared resource, causing again a _race condition_. 
+
+But how to solve this issue? We need a real atomic operation that has to be used within the acquire loop. There are different solutions that we can adopt (for example use the `xchgl` instruction in the for loop, that is guaranteed to be atomic), but as usual we want to keep thing extremely simple, and in this case we can avail of the help of the compiler (gcc) that provides a whole set of atomic functions to be used by our kernel.
```
DeanoBurrito, line 161: I wonder if it's worth adding a sentence here explaining what we mean by atomic?

04_Scheduling/04_Locks.md
dreamos82: yeah sorry leftover from previous example XD

04_Scheduling/04_Locks.md
dreamos82: ahaha the curse of copying and pasting XD

04_Scheduling/03_Processes_And_Threads.md
```diff
@@ -169,94 +169,165 @@ Now any further operations on this file can use the returned id to reference thi
...
+    process->pid = next_process_id++;
+    process->threads = NULL;
+    process->root_page_table = vmm_create();
+    strncpy(process->name, name, NAME_MAX_LEN);
+    return process;
+}
 ```
 
-We need to update also the `process_t` data structure with few changes: 
+The `alloc_process` function is just a placeholder, for brevity, `vmm_create` is also a placeholder. Replace these with your own code.
```
dreamos82, line 271: even if iti quite clear just spending a sentence to explain what are the two function expected output, especially for the vmm_create can be useful?

04_Scheduling/03_Processes_And_Threads.md
```diff
@@ -0,0 +1,385 @@
...
+    process_t* process = alloc_process();
+    process->pid = next_process_id++;
+    process->threads = NULL;
+    process->root_page_table = vmm_create();
+    strncpy(process->name, name, NAME_MAX_LEN);
+    return process;
+}
+```
+
+The `alloc_process` function is just a placeholder, for brevity, `vmm_create` is also a placeholder. Replace these with your own code.
```
dreamos82, line 271: even if iti quite clear just spending a sentence to explain what are the two function expected output, especially for the vmm_create can be useful?

04_Scheduling/03_Processes_And_Threads.md
```diff
@@ -169,94 +169,165 @@ Now any further operations on this file can use the returned id to reference thi
...
+
+We will also need a new status for the thread:  `SLEEPING`.
+
+To actually put a thread to sleep, we'd need to do the following:
+
+- Change the thread's status to `SLEEPING`. Now the scheduler will not run it since it's not in the `READY` state.
+- Set the `wake_time` variable to the requested amount of time in the future.
+- Force the scheduler to change tasks, so that the sleep function does not immediately return, and then sleep on the next task switch.
 
+We will need to modify the scheduler check the wake time of any sleeping threads it encounters. If the wake time is in the past, then we can change the thread's state back to `READY`.
```
dreamos82, line 303: we will need to modify the schedueler TO check?

04_Scheduling/03_Processes_And_Threads.md
```diff
@@ -169,94 +169,165 @@ Now any further operations on this file can use the returned id to reference thi
...
 
-The scheduler needs to be updated too, the main change is thata now the iret_frame (the process context) is no longer in the `process_t` structure, but now is a field within `threads` field. So we need to update the function to read it from there. 
+At a first glance this may seem like the same thing, but it eliminates unnecessary timer interrupts, when no task switch is occuring. It also removes the idea of a `quantum`, since you can run a thread for any arbitrary amount of time, rather than a number of timer intervals.
 
-These are most of the changes needed to start to use threads, even if in our case we are allowing a single thread per process, moving towards mulitple threads per process is pretty easy now, we just need eventually to make them into a linked list, adding a `thread_t *next` field to the data structure (or using a fixed length array like we did in the process scheduler), and update our scheduler to iterate through the threads within a process. 
+*Authors note: Tickless schedulers are usually seens as more accurrate and operate with less latency than periodic ones, but this comes at the cost of added complexity.*
```
dreamos82, line 324: seens?

04_Scheduling/03_Processes_And_Threads.md
```diff
@@ -0,0 +1,385 @@
...
+
+We will also need a new status for the thread:  `SLEEPING`.
+
+To actually put a thread to sleep, we'd need to do the following:
+
+- Change the thread's status to `SLEEPING`. Now the scheduler will not run it since it's not in the `READY` state.
+- Set the `wake_time` variable to the requested amount of time in the future.
+- Force the scheduler to change tasks, so that the sleep function does not immediately return, and then sleep on the next task switch.
+
+We will need to modify the scheduler check the wake time of any sleeping threads it encounters. If the wake time is in the past, then we can change the thread's state back to `READY`.
```
dreamos82, line 303: we will need to modify the schedueler TO check?

04_Scheduling/03_Processes_And_Threads.md
```diff
@@ -0,0 +1,385 @@
...
+
+### Advanced Designs
+
+We've discussed the common approach to writing a scheduler using a periodic timer. There is another more advanced design: the tickless scheduler. While we won't implement this here, it's worth being aware of. 
+
+The main difference is how the scheduler interacts with the timer. A periodic scheduler tells the timer to trigger at a fixed interval, and runs in response to the timer interrupt. A tickless scheduler instead uses a one-shot timer, and set the timer to send an interrupt when the next task switch is due.
+
+At a first glance this may seem like the same thing, but it eliminates unnecessary timer interrupts, when no task switch is occuring. It also removes the idea of a `quantum`, since you can run a thread for any arbitrary amount of time, rather than a number of timer intervals.
+
+*Authors note: Tickless schedulers are usually seens as more accurrate and operate with less latency than periodic ones, but this comes at the cost of added complexity.*
```
dreamos82, line 324: seens

04_Scheduling/03_Processes_And_Threads.md
```diff
@@ -0,0 +1,379 @@
...
+} process_t;
+```
+
+Define `NAME_MAX_LEN` to be a length of your choosing, a good starting place is 64. We've taken the approach of storing the name inside of the control block, but you could also store a pointer to a string on the heap. Using the heap would require more care when using this struct, as you'd have to be sure you managed the memory properly. 
+
+How do we assign pids? We're using to use a bump allocator: which if you'll remember is just a pointer that increases. The next section covers the details of this. It's worth noting that since we're on a 64-bit architecture and using `size_t`, we don't really have to worry about overflowing this simple allocator, as we have 18446744073709551615 possible ids. That's a lot!
+
+### Creating A New Process
+
+Creating a process is pretty trivial. We need a place to store the new `process_t` struct, in our case the static array, but you might have another data struct for it. We'll want a new function that creates a new process for us. We're going to need the starting address for the new process, and it can be nice to in an argument to this function.
```
dreamos82, line 60: near the end of the second sentence: 

and it can be nice to in an argument

something missing again! :)

04_Scheduling/03_Processes_And_Threads.md
```diff
@@ -0,0 +1,379 @@
...
+
+Copying the higher half page tables like this can introduce a subtle issue: If the kernel modifies a pml4 entry in one process the changes won't be visible in any of the other processes. Let's say the kernel heap expands across a 512 GiB boundary, this would modify the next pml4 (since each pml4 entry is responsible for 512 GiB of address space). The current process would be able to see the new part of the heap, but upon switching processes the kernel could fault when trying to access this memory.
+
+While we're not going to implement a solution to this, but it's worth being aware of. One possible solution is to keep track of the current 'generation' of the kernel pml4 entries. Everytime a kernel pml4 is modified the generation number is increased, and whenever a new processes is loaded it's kernel pml4 generation is checked against the current generation. If the current generation is higher, we copy it's kernel tables over, and now the page tables in are synchronized again.
```
dreamos82, line 121: at the end of this paragraph: 

and now the page tables in are synchronized again

i think in is not needed here.

04_Scheduling/03_Processes_And_Threads.md
```diff
@@ -0,0 +1,379 @@
...
+
+Don't forget to load the new process's page tables before leaving the `schedule()`.
+
+#### The Heap
+
+Let's talk about the heap for a moment. With each process being isolated, they can't really share any data, meaning they can't share a heap, and will need to bring their own. The way this usually works is programs link with a standard library, which includes a heap allocator. This heap is exposed through the familiar `malloc()`/`free()` functions, but behind the scenes this heap is calling the VMM and asking for more memory when needed. 
+
+Of course the kernel is the exception, because it doesn't live in it's own process, but instead lives in *every* process. It's heap is available in every process, but can only be used by the kernel.
```
dreamos82, line 129: It's or its?

04_Scheduling/03_Processes_And_Threads.md
```diff
@@ -0,0 +1,379 @@
...
+
+What this means is when we look at loading programs in userspace, these programs will need to provide their own heap. However we're only running threads within the kernel right now, so we can just use the kernel heap.
+
+### Resources
+
+Resources are typically implemented an opaque handle: a resource is given an id by the subsystem it interacts with, and that id is used to represent the resource outside of the subsystem. Other kernel subsystems or programs can use this id to perform operations with the resource. These resources are usually tracked per process.
```
dreamos82, line 135: Resources are typically implemented ... an opaque handle

something is missing in the sentence

04_Scheduling/03_Processes_And_Threads.md
```diff
@@ -0,0 +1,379 @@
...
+
+What this means is when we look at loading programs in userspace, these programs will need to provide their own heap. However we're only running threads within the kernel right now, so we can just use the kernel heap.
+
+### Resources
+
+Resources are typically implemented an opaque handle: a resource is given an id by the subsystem it interacts with, and that id is used to represent the resource outside of the subsystem. Other kernel subsystems or programs can use this id to perform operations with the resource. These resources are usually tracked per process. 
+
+As an example, let's look at opening a file. We wont go over the code for this, as it's beyond the scope of this chapter, but it serves as a familiar example.
+
+When a program goes to open a file, it asks the kernel's VFS (virtual file system) to locate a file by name. Assuming the file exists and can be accesses, the VFS loads the file into memory and keeps track of the buffer holding the loaded file. Let's say this is the 23rd file the VFS has opened, it might be assigned the id 23. You could simply use this id as your resource id, however that is a system-wide id, and not specific to the current process.
```
dreamos82, line 139: can be accesses --> can beaccessed

04_Scheduling/03_Processes_And_Threads.md
```diff
@@ -0,0 +1,379 @@
...
+As an example, let's look at opening a file. We wont go over the code for this, as it's beyond the scope of this chapter, but it serves as a familiar example.
+
+When a program goes to open a file, it asks the kernel's VFS (virtual file system) to locate a file by name. Assuming the file exists and can be accesses, the VFS loads the file into memory and keeps track of the buffer holding the loaded file. Let's say this is the 23rd file the VFS has opened, it might be assigned the id 23. You could simply use this id as your resource id, however that is a system-wide id, and not specific to the current process. 
+
+Commonly each process holds a table that maps process-specific resource ids to system resource ids. A simple example would be an array, which might look like the following:
+
+```c
+#define MAX_RESOURCE_IDS 255
+typedef struct {
+... other fields ...
```
dreamos82, line 146: we should check how is rendered, maybe better to leave these sentences with comment format to have a better rendering...

04_Scheduling/04_Locks.md
```diff
@@ -0,0 +1,201 @@
...
+
+## Introduction
+
+Now that we have a scheduler, we can run multiple threads at the same time. This introduces a new problem though: shared resources and synchronization. 
+
+Imagine we have a shared resource that can be accessed at a specific address. This resource could be anything from MMIO, a buffer or some variable, the important part is that multiple threads *can* access it at the same time. 
+
+For our example we're going to say this resouce is a NS16550 uart at address `0xDEADBEEF`. If you're not familiar with this type of uart device, it's the de facto standard for serial devices. The COM ports on x86 use one of these, as do many other platforms. 
+
+The key things to know are that if you write a byte at that address, it will be sent over the serial port to whatever is one the other send. So if you want to send a message, you must send it one character at a time, at the address specified (`0xDEADBEEF`).
```
dreamos82, line 11: on the other send?

04_Scheduling/04_Locks.md
```diff
@@ -0,0 +1,201 @@
...
+
+![Shared Resource Sequence](Images/sharedressequence.png)
+
+What we'd expect to see is one of two outcomes: `I am the first string while I am the second` or `while I am the second I am the first string`.
+
+This situation we've described is an example of a _race condition_. The order of accesses the the shared resource (the uart) matters here, so we need a way to protect it. This where a lock comes in.
+
+## Implementing A Lock
+
+A _lock_ provides us with something called mutual exclusion: only one thread can hold the lock at a time, while the rest must wait for the lock to be free again before they can take. While a thread is holding the lock, it's allowed to access the shared resource.
```
dreamos82, line 59: not sure if something is missing at the end of the first sentence: 

while the rest must wait for the lock to be free again before they can take...

04_Scheduling/04_Locks.md
```diff
@@ -0,0 +1,201 @@
...
+
+This ensures that each call to `serial_log` completes before another is allowed to start, meaning each message is written properly without being jumbled by another.
+
+It's worth noting that each instance of a lock is independent, so to protect a single resource, you must use a single lock.
+
+## First Implementation
+
+Let's look at how `acquire` and `release` should function:
+
+- `acquire` will wait until the lock if free, and then take it. This prevents a thread from going any further than this function until it has the lock.
```
dreamos82, line 103: if -> is

04_Scheduling/04_Locks.md
```diff
@@ -0,0 +1,201 @@
...
+
+Using the compiler built-in functions like this ensures that the compiler will always generate the correct atomic instructions for us. These are also cross-platform: meaning if you port your kernel to another cpu architecture, you can still use these functions.
+
+### Side Effects
+
+While atomic operations are fantastic, they are often quite slow compared to their non-atomic counterparts. Of course it's impossible to implement a proper lock without atomics so we must use them, however it's equally important not to *over use* them.
+
+## Locks and Interrupts
+
+With the introduction of locks we've also introduced a new proble.
```
dreamos82, line 182: proble...m

04_Scheduling/03_Processes_And_Threads.md
```diff
@@ -0,0 +1,379 @@
...
+```
+
+While this is functional, there's a few problems:
+
+- They can't be easily identified, how do we know the difference between procesess.
+- All processes currently share the same address space, and as a byproduct, the same virtual memory allocator. 
+- We can't keep track of any resources they might be using, like file handles or network sockets.
+- We can't prioritize them, as we dont know which ones are more important.
+
+We're not going to look at how to solve all of these, but we'll the important ones.
```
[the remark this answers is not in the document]
DeanoBurrito: haha yes, something was missing.

04_Scheduling/03_Processes_And_Threads.md
DeanoBurrito: Ooh I didn't think about that, good idea. I will stick to comments.

I've done this in a few other places too I think, so I'll try to find those and fix them.

PR #31, opened 2022-07-06
Updated Interrupts and GDT chapters

DeanoBurrito:


GDT.md
```diff
@@ -214,56 +229,52 @@ A more complex example of a GDT is the one used by the stivale2 boot protocol:
...
+void flush_gdt()
+{
+    asm volatile("\
+                mov $0x10, %ax \n\
+                mov %ax, %ds \n\
+                mov %ax, %es \n\
+                mov %ax, %fs \n\
+                mov %ax, %gs \n\
+                mov %ax, %ss \n\
+                \n\
```
dreamos82, line 269: just double checking: is this empty line intentional? if yes just ignore leave as it is. 

Why the extra \ at the end of each line?

InterruptHandling.md
```diff
@@ -1,146 +1,400 @@
-# IDT
+# Interrupt Handling on x86
```
dreamos82, line 1: i would say x86_64?

InterruptHandling.md
```diff
@@ -1,146 +1,400 @@
...
 
-Although the IDT in long mode is pretty similar to the IA32 one, there are few differences. I'll try to cover them in this document.
+As the title implies, this chapter is purely focused on x86. Other platforms will have different mechanisms for handling interrupts.
 
-## IDT Entry Structure
+If you're not familiar with the term *interrupt*, it's a way for the cpu to tell your code that something unexpected or unpredictable has happened, and that you need to handle it. When an interrupt is triggered, the cpu will *serve* the interrupt by loading the *interrupt handler* we specified. The interrupt handler itself is just a function, but with a few special conditions. 
 
-First important difference, is that while the IA32 idt entry is 64 bit, in long mode is 128 bit, and the structure is little bit different. 
+Interrupts get their name because they interrupt the normal flow of execution, instead stopping whatever code was running on the cpu, running a handler function, and then resuming the previously running code. Interrupts can signal a number of events from the system, from fatal errors to a device telling you it has some data for you to read.
```
dreamos82, line 7: that instead looks out of place?

InterruptHandling.md
```diff
@@ -1,146 +1,400 @@
...
+There will be situations where you don't want to be interrupted, usually in some kind of critical section. In this case, x86 actually provides a flag you can use to disable almost all interrupts. Bit 9 of the `flags` register is the interrupt flag, and like other flag bits it has dedicated instructions for clearing/setting it:
+
+- `cli`: Clears the interrupt flag, preventing the cpu from serving interrupts.
+- `sti`: Sets the interrupt flag, letting the cpu serve interrupts.
+
+### Non-Maskable Interrupts
+
+When the interrupt flag is cleared, most interrupts will be *masked* meaning they will not be served. There is a special case where an interrupt will still be served by the cpu: the *non-maskable interrupt* or NMI. These are extremely rare, and often a result of a critical hardware failure, therefore it's perfectable acceptable to simply have your operating system panic in this case. 
+
+*Authors note: Don't let NMIs scare you, I've never run actually run into one on real hardware. You do need to be aware that they exist and can happen at any time, regardless of the interrupt flag.*
```
dreamos82, line 22: two run?

InterruptHandling.md
```diff
@@ -1,146 +1,400 @@
...
+
+## Setting Up For Handling Interrupts
+
+Now we know the theory behind interrupts, let's take a look at how we interact with them on x86. As expected, it's a descriptor table! We will be referencing some GDT selectors in this, so you'll need to have your own GDT setup. We're also going to introduce a few new terms:
+
+- Interrupt Descriptor: A single entry within the interrupt descriptor *table*, it describes what the cpu should do when a specific interrupt occurs.
+- Interrupt Descriptor Table: An array of interrupt descriptors., usually referred to as the IDT.
```
dreamos82, line 29: when dot and comma can´t stay on the same column XD

InterruptHandling.md
```diff
@@ -1,146 +1,400 @@
...
+    uint8_t ist;
+    uint8_t flags;
+    uint16_t address_mid;
+    uint32_t address_high;
+    uint32_t reserved;
+} __attribute__((packed));
 ```
 
-Where: 
+Note the use of the packed attribute! Since this structure is processed by hardware, we dont want the compiler to insert any padding in our struct, we want it to look exactly as we defined it.
```
dreamos82, line 54: could be useful to mention the fact that the idt entry size is 128 bit  ?

GDT.md
DeanoBurrito: It tells the compiler that the string doesn't end at the end of the line.
Empty line is for clarity.

InterruptHandling.md
```diff
@@ -1,146 +1,400 @@
...
+
+### An Example Stub
+
+Armed with the above infomation, you should be able to implement your own handler stubs. One common way to do this is using an assembler macro. Here you would create one macro that pushes all registers, calls a C function and then pops all registers before executing `iret`. What about the optional error code? Well, the easiest solution is to define *two* macros, one like the previous one, and another that pushes a pretend error code of 0, before pushing all the general registers. Because we know which vectors push an error code and which don't, we can change which macro we use. 
+
+The benefit of this is your stack will always look the same regardless of whether a real error was used or not. This allows us to do all sorts of things later on.
+
+Another solution, is to only write a single assembly stub like the first macro. Then for each handler function you could either just jump to the stub function (if an error code was pushed by the cpu), or push a dummy error code and then jumpo to the stub function. We'll go with the second option.
```
dreamos82, line 151: jumpo...  lol

InterruptHandling.md
```diff
@@ -1,146 +1,400 @@
...
-Otherwise when the cpu receives an interrupt, the halt flag will be cleared in order to execute the handler, 
-and isn't restored. Therefore when you exit the interrupt, if not inside a loop, you'll start executing junk data after the halt!
+```x86asm
+interrupt_stub:
+push %rax
+push %rbx
+[ ... push other registers here ... ]
+push %r14
+push %r15
+
```
dreamos82, line 162: should you add the vecttor number and error code in the stub?

InterruptHandling.md
```diff
@@ -1,146 +1,400 @@
...
+```c
+extern char vector_0_handler[];
+
+for (size_t i = 0; i < 256; i++)
+    set_idt_entry(i, (uint64_t)vector_0_handler + (i * 16), 0);
+```
+
+### Sending EOI
+
+With that done, we can now enter and return from interrupt handlers correctly! You should keep in mind that this is handling interrupts with the cpu. The cpu usually does not send interrupts to itself, it receives them from an external device like the local APIC. APICs are discussed in their own chapter, but you will need to tell the local APIC that you haven handled the latest interrupt. This is called sending the EOI (End Of Interrupt) signal.
```
dreamos82, line 223: haven

InterruptHandling.md
```diff
@@ -1,146 +1,400 @@
...
+- `%rip`: The previous instruction pointer.
+
+Optionally, for some vectors the cpu will push a 64-bit error code (see the table below for specifics).
+This structure is known as an *iret frame*, because to return from an interrupt we use the `iret` instruction, which pops those five values from the stack. Hopefully the flow of things is clear at this point: the cpu serves the interrupt, pushes those five values onto the stack. Our handler function runs, and then executes the `iret` instruction to pop the previously pushed values off the stack, and return to the interrupted code.
+
+### An Example Stub
+
+Armed with the above infomation, you should be able to implement your own handler stubs. One common way to do this is using an assembler macro. Here you would create one macro that pushes all registers, calls a C function and then pops all registers before executing `iret`. What about the optional error code? Well, the easiest solution is to define *two* macros, one like the previous one, and another that pushes a pretend error code of 0, before pushing all the general registers. Because we know which vectors push an error code and which don't, we can change which macro we use.
```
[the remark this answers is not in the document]
DeanoBurrito: I'm not sure what there would be to explain? You simply can't write that kind of code in C. Inline assembly does not guarantee the compiler won't interfere with the instructions (it may reorder them, or add extras). The only way to write an interrupt handler in C is the use the interrupt attribute, but then you can't access the saved registers, and I dont really want to suggest that.

InterruptHandling.md
DeanoBurrito: lol:D

PR #34, opened 2022-10-13
Chapter 6: IPC

DeanoBurrito:


06_IPC/02_Shared_Memory.md
```diff
@@ -0,0 +1,94 @@
...
+
+## IPC Manager
+
+We're going to implement an *IPC manager* to keep track of shared memory.
+
+At it's heart, our IPC manager is going to be a list of physical memory ranges, with a name attached to each range. When we say physical memory *range*, this just refers to a number of physical pages located one after the other (i.e. contiguous). Attaching a name to a range lets us identity it, and we can even give some of these names special meanings, like `/dev/stdout` for example. In reality this is not how stdout is usually implemented, but it serves to get the point across.
```
dreamos82, line 21: identity or identify?

06_IPC/02_Shared_Memory.md
```diff
@@ -0,0 +1,202 @@
...
+
+### Potential Issues
+
+There's a few potential issues to be aware of with shared memory. The biggest one is to be careful when writing data that contains pointers. Since each process interacting with shared memory may see the shared physical memory at a different virtual address, any pointers you write here may not be valid.
+
+Best practice is to store data *relative to the base*, this way each process can read the pointer from the shared memory, and add it's own virtual offset.
+
+Another problem that may arise is your compile optimizing away reads and writes to the shared memory. This can happen because the compiler sees these memory accesses are having no effect on the rest of the program. This is the same issue you might have experienced with MMIO (memory mapped io) devices, and the solution is the same: make any reads or write `volatile`.
```
dreamos82, line 135: compile -> compiler

06_IPC/02_Shared_Memory.md
```diff
@@ -0,0 +1,202 @@
...
+    if (flags & VMM_FLAG_SHARED_MEMORY)
+        free_shared_memory(phys_addr);
+    else
+        pmm_free_pages(phys_addr, length / PAGE_SIZE);
+
+    //do other vmm free stuff, like adjusting page tables.
+}
+```
+
+The `vmm_get_Flags` is a made up function, it just returns the flags used for a particular virtual memory allocation, we also use a function to manually walk the page tables and get the physical address mapped to this virtual address (`get_phys_addr`). For details on how to get the physical address mapped to a virtual one, see the section on paging.
```
dreamos82, line 157: vmm_get_Flags -> vmm_get_flags

06_IPC/02_Shared_Memory.md
```diff
@@ -0,0 +1,202 @@
...
+            list = found->next;
+        else
+            prev->next = found->next;
+        free(found);
+    }
+    release(list_lock);
+}
+```
+
+Again we're omitted error handling and checking for `NULL` to keep the examples concise, but you should handle these cases in your code. The first part of the function should look similar, it's the same code used in `access_shared_memory`. The interesting part happens when the reference count reaches zero: we free the physical pages used, and free any memory we previously allocated. We also remove the shared memory instance from the linked list.
```
dreamos82, line 190: we're -> we've?

06_IPC/02_Shared_Memory.md
```diff
@@ -0,0 +1,202 @@
...
+
+## Interesting Applications
+
+Most applications will default to using message passing (in the form of a pipe) for their IPC, but shared memory is a very simple and powerful alternative. It's biggest advantage is that no intervention from the kernel is required during runtime: multiple processes can exchange data at their own pace, quite often faster than you could with message passing, as there are less context switches to the kernel.
+
+More commonly a hybrid approach is taken, where processes will write into shared memory, and if a receiving process doesn't check it for long enough, the sending process will invoke the kernel to send a message to the receiving process.
+
+## Access Protection
+
+It's important to keep in mind that we have no access protection in this example. Any process can view any shared memory if it knows the correct name. This is quite unsafe, especially if you're exechanging sensitive information this way. Commonly shared memory is presented to user processes through the virtual file system (we'll look at tihs more later), this has the benefit of being able to use the access controls of the VFS for protecting shared memory as well.
```
dreamos82, line 200: exechanging -> exchanging

02_Memory_Management/04_Virtual_Memory_Manager.md
```diff
@@ -1,4 +1,203 @@
...
+
+```c
+void* vmm_alloc(size_t length, size_t flags, void* arg);
+```
+
+The `length` and `flag` fields are how many bytes we want, and the type of memory we want. The `void* arg` field is an oddity, it's a place to pass in extra data if the allocation needs it. We'll see why this is useful later.
+
+The function will try to find a space in virtual memory big enough to hold `length` bytes, if it can it marks this space as taken, and returns the base address. If you're using immediate-backing (as opposed to demand-backing) this is where you would allocate pages from the pmm, and modify the page tables so this physical memory is available.
+
+An example implementation is below, remember we're using a sorted linked list for storing virtual memory objects.
```
dreamos82, line 99: i would split if possible the introduction of the vmm alloc in several parts, otherwise we are just giving code ready to copy and paste to the reader. The idea is that the reader should be able to implement it by himself reading our guide, not us giving him the code ready.

06_IPC/03_Message_Passing.md
```diff
@@ -0,0 +1,176 @@
...
+You may wish to use the VFS and file descriptors in your own design, or something completely different! Several of the ideas discussed in the previous chapter on shared memory apply here as well, like access control. We won't go over those again, but they're worth keeping in mind here.
+
+## How It Works
+
+After the initial setup, the implementation of message passing is similar to a relay race:
+
+- Process 1 wants to receive incoming messages on an endpoint, so it calls a function telling the kernel it's ready. This function will only return once a flag has been set on the endpoint that a message is ready, and otherwise blocks the thread. We'll call this function `ipc_receive()`.
+- Some time later process 2 wants to send a message, so it allocates a buffer and writes some data there.
+- Process 2 now calls a function to tell the kernel it wants to send this buffer as a message to an endpoint. We'll call this function `ipc_send()`.
+- Inside `ipc_send()` the buffer is copied into kernel memory. In our example we'll use the heap for this memory. We can then set a flag on the endpoint telling it that has a message has been received.
```
dreamos82, line 22: typo? that has ha message has been...

06_IPC/03_Message_Passing.md
```diff
@@ -0,0 +1,176 @@
...
+- The next time process 1 runs, `ipc_receive()` will see that the flag has been set, and copy the message from the kernel buffer into a buffer for the program.
+- The `ipc_receive()` function can also free the kernel buffer, before returning and letting process 1 continue as normal.
+
+What we've described here is a double-copy implementation of message: because the data is first copied into the kernel, and then out of it. Hence we performed two copy operations.
+
+## Initial Setup
+
+As mentioned, there's some initial setup that goes into message passing: we need to create an endpoint. We're going to need a way to identify each endpoint, so we'll use a string. We'll also need a way to indicate when a message is available, the address of the buffer containing the message, and the length of the message buffer.
+
+Since the struct representing the endpoint is going to be accessed by multiple proceses, we'll want a lock to protect the data from race conditions. For we'll use the following struct to represent our endpoint:
```
dreamos82, line 33: For we'll use, i think something is missing?

06_IPC/03_Message_Passing.md
```diff
@@ -0,0 +1,176 @@
...
+    void* msg_buffer;
+    size_t msg_length;
+    spinlock_t msg_lock;
+    ipc_endpoint* next;
+};
+```
+
+To save some space we'll use `NULL` as the message address to represent that there is no message available. 
+
+If you're wondering about the `next` field, that's because we're going to store these are a linked list. You'll want a variable to store the head of the list, and a lock to protect the list anytime it's modified.
```
dreamos82, line 47: typoe? we're going to store these are a linked list...

06_IPC/03_Message_Passing.md
```diff
@@ -0,0 +1,176 @@
...
+        ipc_endpoint* last = first_endpoint;
+        while (last->next != NULL)
+            last = last->next;
+        last->next = ep;
+    }
+    release(&endpoints_lock);
+}
+```
+
+As you can see creating a new endpoint is pretty simple, and most of the code in the exaple function is actually for managing the linked list.
```
dreamos82, line 78: exaple -> example

06_IPC/03_Message_Passing.md
```diff
@@ -0,0 +1,176 @@
...
+
+Now our endpoint has been added ot the list! As always we omitted checking for errors, and we didn't check if an endpoint with this name already exists. In the real world you'll want to handle this things.
```
dreamos82, line 80: ot -> to

06_IPC/03_Message_Passing.md
```diff
@@ -0,0 +1,176 @@
...
+```c
+void* msg_copy = malloc(length);
+memcpy(msg_copy, buffer, length);
+```
+
+Why do we make a copy of the original message? Well if we dont, the sending process has to keep the original message around until the receiver has processed it. We don't have a way for the receiver to communicate that it's finished reading the message. By making a copy, we can return from `ipc_send()` as soon as the message is sent, regardless of when the message is read. Now the sending process is free to do what it wants with the memory holding the original message as soon as `ipc_send()` has completed.
+
+If you're performing this IPC as part of a system call from userspace, the memory containing the original message is unlikely to be mapped in the receiver's address space anyway, so we have to copy it into the kernel's address space, which is mapped in both processes.
+
+All that's left is to tell the receiver it has a message is available by placing the buffer address on the endpoint. Again, notice the use of the lock to prevent race conditions while we mess with the internals of the endpoint.
```
dreamos82, line 128: it has a message is available

that "is" is really necessary?

06_IPC/03_Message_Passing.md
```diff
@@ -0,0 +1,176 @@
...
+
+```c
+acquire(&target->lock);
+target->msg_buffer = msg_copy;
+target->msg_length = length;
+release(&target->lock);
+```
+
+After the lock on the endpoint is released, the sent has been sent! Now it's up to the receiving thread to check the endpoint and set the buffer to NULL again.
```
dreamos82, line 137: the sent has been sent? 

pun intended or typo?

06_IPC/03_Message_Passing.md
```diff
@@ -0,0 +1,176 @@
...
+- Allow for multiple messages to be stored on an endpoint.
+- Fail to send the message, instead returning an error code from `ipc_send()`.
+
+The first option is recommended, as it's likely there will be some processes that handle a lot of messages. Implementing this is left as an exercise to the user, but a simple implemenation might use a struct to hold each message (the buffer address and length) and a next field. Yes, more linked lists!
+
+Sending messages would now mean appending to the list instead of writing the buffer address as before. 
+
+## Receiving
+
+We can send messages, next let's look at receiving them. We're going to use a basic (and inefficient) example, but it shows how it could be done.
```
dreamos82, line 152: shouldn't be better adding now? Like:

"We have seen how to send messages, now let's look at how to receive them" (or something similar)

The current sentence give a sense of rush...

06_IPC/03_Message_Passing.md
```diff
@@ -0,0 +1,176 @@
...
+endpoint->msg_data = NULL;
+endpoint->msg_length = 0;
+release(&endpoint->lock)'
+```
+
+At this point the endpoint is now ready to receive another message, and we've got a copy of the message in `local_copy`. You're successfully passed a message from one address space to another!
+
+## Additional Notes
+
+- We've described a double-copy implementation here, but you might want to try a single-copy implemenation. Single-copy implementations *can* be faster, but they require extra logic. For example the kernel will need to access the recipient's address space from the sender's address space, how do you manage this? If you have all of physical memory mapped somewhere (like an identity map) you could use this, otherwise you will need some way to access this memory.
```
dreamos82, line 174: not sure what's going on here, but the sintax from the github syntax viewer is strange, try to double check if there are some markdown issues.

06_IPC/03_Message_Passing.md
```diff
@@ -0,0 +1,176 @@
...
+- A processing waiting on an endpoint (to either send or receive a message) could be waiting quite a while in some circumstances. This is time the cpu could be doing work instead of blocking and spinning on a lock. A simple optimization would be to put the thread to sleep, and have it be woken up whenever the endpoint is updated: a new message is sent, or the current message is read.
```
dreamos82, line 175: processing? maybe you mean procesS?

dreamos82: and same error here too with the syntax higlighter.

06_IPC/03_Message_Passing.md
```diff
@@ -0,0 +1,176 @@
...
+- In this example we've allowed for messages of any size to be sent to an endpoint, but you may want to set a maximum message size for each endpoint when creating it. This makes it easier to receive messages as you know the maximum possible size the message can be, and can allocate a buffer without checking the size of the message. This might seem silly, but when receiving a message from userspace the program has to make a system call each time it want's the kernel to do something. Having a maximum size allows for one-less system call. Enforcing a maximum size for messages also has security benefits.
```
dreamos82, line 176: the syntax prbolems stops after "In this example we`"

dreamos82: it want's ?

dreamos82 (changes requested):

06_IPC/02_Shared_Memory.md
```diff
@@ -0,0 +1,204 @@
...
+
+The solution we're going to use is *reference counting*. Everytime a VMM maps shared memory into it's page tables, we increase the count by 1. Whenever a VMM exits and goes to free physical memory, it will first check if it's shared memory or not. If it's not, it can be freed as normal, but if it's shared memory we simply decrease the reference count by 1. Whenever decrementing the reference count, we check if it's zero, and only then do we free the physical memory. The reference count can be thought of as a number that represents how many people are using a particular shared memory instance, with zero meaning no-one is using it. If no-one is using it, we can safely free it.
```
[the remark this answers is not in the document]
DeanoBurrito: I'm not sure how to how to describe it, but putting the 'do' is a way of adding emphasis. Not necessary though, and just clutters the sentence, removed it :)

06_IPC/02_Shared_Memory.md
```diff
@@ -0,0 +1,204 @@
...
+    tail->next = shared_mem;
+    release(list_lock);
+
+    return shared_mem->physical_base;
+}
+```
+
+This code is the core to implementing shared memory, it allows us to create a new shared memory region and gives us it's physical address. The VMM can then map this physical address into the memory space of the process like it would do with any other memory.
+
+Notice the use of the lock functions (`acquire` and `release`) when we access the linked list. Since this single list is shared between all processes that use shared memory, we have to protect it so we accidentally corrupt it. This is discussed further in the chapter on scheduling.
```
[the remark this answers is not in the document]
DeanoBurrito: haha yeah I forgot the word 'dont'. We dont want to corrupt it.

06_IPC/03_Message_Passing.md
DeanoBurrito: type :)

06_IPC/03_Message_Passing.md
DeanoBurrito: yeah thats fair, I think I was rushing to get this done before getting off the train haha. I went with your version.

06_IPC/03_Message_Passing.md
DeanoBurrito: It was a misplaced single quote (') instead of a semicolon on the previous code block :)

06_IPC/02_Shared_Memory.md
```diff
@@ -81,7 +81,7 @@ void* create_shared_memory(size_t length, const char* name) {
 
 This code is the core to implementing shared memory, it allows us to create a new shared memory region and gives us it's physical address. The VMM can then map this physical address into the memory space of the process like it would do with any other memory.
 
-Notice the use of the lock functions (`acquire` and `release`) when we access the linked list. Since this single list is shared between all processes that use shared memory, we have to protect it so we accidentally corrupt it. This is discussed further in the chapter on scheduling.
+Notice the use of the lock functions (`acquire` and `release`) when we access the linked list. Since this single list is shared between all processes that use shared memory, we have to protect it so we dont accidentally corrupt it. This is discussed further in the chapter on scheduling.
```
dreamos82, line 84: dont or don´t ?

dreamos82 (approved):

PR #35, opened 2022-10-30
File system

dreamos82:
Currently draft state


07_VirtualFileSystem/01_README.md
```diff
@@ -0,0 +1,35 @@
...
+It will be divided into two main topics: 
+
+* The Virtual File System (VFS): is a layer that works as an abstraction for the different file systems supported by our os, in this way the application and the kernel use a unified interface of functions to interact with files and directories on different fs, it offers functions like `open, write, opendir`
+* In the File System section we will how they work within a VFS interface, and implement a very basic one to use within our OS,
+
+## A quick recap
+
+Before proceeding is useful to recap some basic file system concepts.
+
+The main purpose of a file system is to store data and make it easily accessible on a human readable way. What a file system does is organize how the data is stored, how they are represented on the disk, and provide functionalities to access, create, update and delete them. More advanced FS can also provide some kind of recovery mechanism (aka `journaling`) but we are not going to cover it because it's out of the scope of this guide.
```
DeanoBurrito, line 22: Nice, I would probably mention that a filesystem can provide permissions oto.

07_VirtualFileSystem/01_README.md
```diff
@@ -0,0 +1,35 @@
...
+
+Any file system is provided by a driver, so an OS has to implement one for each type they want to support (the most popular and easy to implement are probably FAT, tar and ext2).
+
+They will internally represents file, directories in different ways, wether they are just an overhead structure on top of the data, or an entry into an array/list. 
+
+But how to make different file systems with different file representation, be seen in a uniform way by the os? This is achieved through an abstraction layer, the Virtual File System, it will be responsible to talk with the different drivers and provide an unified way to represent them to the user. 
+
+How the different drivers are presented to the user is a design decision, but the two most known (and probably common) way to do that are and of course they have their pros and cons:
+
+* Show them as separated entities, like windows does, where every file system is identified by a unique letter (called the `drive letter`). The pro of this design is that we can have for example same path on two different device with the only difference being the drive letter
```
DeanoBurrito, line 34: Nice contrast, it could be worth mentioning the multi-root is what NT uses, in contrast to unix which uses the single-root style (as you said below).

07_VirtualFileSystem/02_VirtualFileSystem.md
```diff
@@ -114,6 +114,33 @@ void ustar_read(int ustar_fd, void *buffer, size_t count);
...
+
+```c
+int vfs_mount(char *device, char *target, char *fs_type);
+```
+
+Once called it will simply add a new item in the mountpoint on the first available position, and will populate the data mountpoint data structure with the information provided, for example using the array approach, if a free spot is found at index `i` to add a new file sytem we will have something like: 
+
+```c
+mountpoints[i].device = device;
+mountpoints[i].tpye = type;
```
DeanoBurrito, line 129: tpye -> type.

07_VirtualFileSystem/02_VirtualFileSystem.md
```diff
@@ -114,6 +114,33 @@ void ustar_read(int ustar_fd, void *buffer, size_t count);
...
+mountpoints[i].mountpoint = target;
+mountpoints[i[.operations = NULL 
+```
+
+the last line will be populated soon, for now let's leave it to null. 
+
+* The second instruction is for umounting, in this case since we are just unloading the file device from the system, we don't need to know what type is it, so technically we need either the target device or the target folder, the function can actually accept both parameters, but use only one of them, let's call it `vfs_umount`: 
+
+```c
+int vfs_umount(char *device, char *targe);
```
DeanoBurrito, line 139: targe -> target

07_VirtualFileSystem/02_VirtualFileSystem.md
```diff
@@ -0,0 +1,352 @@
...
+```
+
+In this case we need to find the item in the list that contains the required file system, and if found remove it from the list/tree. In our case since we are using an array we need to clear all the fields in the array at the position containing our fs. 
+
+One thing that we should keep in mind is that using an array, once we umount a file system we are just marking that position as available again, so this means that it can be in the middle of the array, and there can be mounted file system after, so we need to keep in mind this while searching for the next free mouuntpoint.
+
+#### A short diversion: the initialization
+
+When the kernel first boot up, of course there is no file system mounted, and all data structures are not allocated, if we decide to use a linked list for example, before the initialization the pointer will point to nothing (or better to garbage) so we will need to allocate the first item of the list to have the first fs accessible.
```
DeanoBurrito, line 148: the phrasing could be improved: "and not all data structures are allocated" is more natural I think.

07_VirtualFileSystem/02_VirtualFileSystem.md
```diff
@@ -0,0 +1,352 @@
...
+
+But where should be the first file system mounted? That again is depending on the project decisions:
+
+* Using a single root approach, the first file system will be mounted on the "/" folder, and this is what we are going to do, this means that all other file systems will be going to stay into subfolders of the root folder. 
+* Using a multi root approach, like windows os, we will have every fs that will have it's own root folder and it will be identified with a letter (A, B, C...)
+* Nothing prevent us to use different approaches, or a mix of them, we can have some file system to share the same root, while some other to have different root, this totally depends on design decision. 
+
+One last thing about the initialization, since our kernel is loaded fully in memory, we don't actually need to have a file system mounted for the kernel to run (even if probably in the future we will need one) so the initialization is just optional to be done during boot time, if the the os already has a kind of shell we can initialize it on the first mount when it will be called.
```
DeanoBurrito, line 158: The first part makes sense, but I can't think of a reason to defer initializing the VFS until later, it's super useful to have a filesystem available, even just an initrd. While technically optional, I wouldn't recommend it.

07_VirtualFileSystem/02_VirtualFileSystem.md
```diff
@@ -0,0 +1,352 @@
...
+
+How to do it is pretty simple, we scan the array, and search for the "longest" mountpoint that is contained in the path, so for in the first example, we can see that there are two items in the array that are contained in the path: "/" (0), and "/home" (3), and the longest one is number 3, so this is what our function is going to return. 
+
+The second path instead has three mountpoints contained into it: "/" (0), "/home/mount" (1), "/home", in this case we are going to return 1. 
+
+The last one, has only one mountpoint that is contained into the path, and it is "/" (0). 
+
+In a single root scenario, there is always at least a mountpoint that is contained into the given path, and this is the root folder "/".
+
+What if for example path doesn't start with a "/"? this means that it's a relative path, it will be explained soon.
```
DeanoBurrito, line 194: It might be nicer for the reader if we include a small sentence explaining how relative paths work, and then we can say we'll look at it more later. At some point a relative path has to become absolute, this can happen in the program calling the vfs itself (meaning the vfs only has to support absolute paths), or you could include a field in the thread control block that stores the current working directory, and use that.

07_VirtualFileSystem/02_VirtualFileSystem.md
```diff
@@ -0,0 +1,352 @@
...
+
+Implementing the function is left as exercise, below we just declare the header (that we will use in the next sections): 
+
+```c
+int get_mounpoint_id(char *path);
+```
+
+If the above function fail it should return a negative number (i.e. -1) to let the caller know that something didn't worked (it should always return at least 0 in a single root implementation).
```
DeanoBurrito, line 202: worked -> work

07_VirtualFileSystem/02_VirtualFileSystem.md
```diff
@@ -0,0 +1,352 @@
...
+
+Why "eventually" ? Having found the mountpoint id for the file doesn't mean that the file exists on that fs, the only thing that exist so far is the mountpoint, but after that the VFS can't really know if the file exists or not, it has to devolve this task to the fs driver, hence it will call the implementation of a function that open a file on that FS that will do the search and return the an error if the file doesn't exists.  
+
+But how to call the fs driver function? Earlier in this chapter when we outlined the `mountpoint_t` structure we added a field called `operations`, of type `fs_operations_t` and left it unimplemented. Now is the time to implement it, this field is going to contain the pointer to the driver functions that will be used by the vfs to open, read, write, and close files:
+
+```c
+struct fs_operations_t {
+	int (*open)(const char *path, int flags, ... );
```
DeanoBurrito, line 290: is the `...` meant to be there haha? Let's keep things simple and the bare minimum.

07_VirtualFileSystem/02_VirtualFileSystem.md
```diff
@@ -0,0 +1,352 @@
...
+	int (*close)(int);
```
DeanoBurrito, line 291: I would say it's worth naming the arguments to these functions, `int` doesn't tell me a whole lot. `int fd` or maybe even `int file_descriptor` is more useful.

07_VirtualFileSystem/02_VirtualFileSystem.md
```diff
@@ -0,0 +1,352 @@
...
+	ssize_t (*write)(int,const void*, size_t);
+};
+
+typedef struct fs_operations_t fs_operations_t;
+```
+
+The basic idea is that once the mountpoint_id has been found (we declared `get_mountpoint_id` in the previous paragraph), the vfs will use the mountpoint item to call the fs driver implementation of the open function, the only thing we need to consider is that when calling the driver function, it doesn't care about the mountpoint part of the path, and if the whole path is passed, we will most likely get an error, since the fs root will start from within the mountpoint folder. What we need thean is a function that get the relative path from an absolute one, the idea and implementation should be pretty straightforward, and just involve few `strcpy` operations, so this is left as exercise, let's just assume that we have one with the following signature: 
+
+```c
+char *get_rel_path(char *mountpoint_part, char* full_path);
```
DeanoBurrito, line 302: This relative path stuff should probably go with the other stuff on relative paths (a few paragraphs earlier).

07_VirtualFileSystem/02_VirtualFileSystem.md
dreamos82: mine was a reference to the current naive kernel we are implementing

07_VirtualFileSystem/02_VirtualFileSystem.md
dreamos82: done

07_VirtualFileSystem/02_VirtualFileSystem.md
```diff
@@ -361,7 +362,45 @@ ssize_t read(int fildes, void *buf, size_t nbyte);
 
 Where the parameters are the opened file descriptor id `fildes`, we want to read, a pointer to a buffer `buf` that will be used to store the read data, and the number of bytes `nbyte` we want to read.
 
-It will return the number of bytes read, and in case of failure -1. Like all other vfs function, what the read will do is search for the file descriptor with id `fildes`, and if it exists call the fs driver function to read data from an opened file. 
+The read function will return the number of bytes read, and in case of failure -1. Like all other vfs function, what the read will do is search for the file descriptor with id `fildes`, and if it exists call the fs driver function to read data from an opened file and fill the `buf` buffer.  
+
+Every time the read is called, it starts from the last byte accessed previously,  so if for example we have a file with the following text: (++++@DT: help me to phrase this part better ++++)
```
DeanoBurrito, line 367: how about:
```
Internally the file descriptor keeps track of a 'read head' (or maybe 'read pointer', either one is fine I think) which points to the last byte that was read. The next read() call will start reading from this byte, before updating the pointer itself.
```

07_VirtualFileSystem/02_VirtualFileSystem.md
```diff
@@ -0,0 +1,433 @@
...
+
+The `close` with signature: 
+
+```c 
+int close(int fildes);
+```
+
+will take a vfs file descriptor as parameter, and will search for it in the opened files list (using an array it will be found at `vfs_opened_files[fildes]`) and if found it should first call the fs driver function to close a file (if present), emptying all data structures associated to that file descriptor (i.e. if there are data on pipes or FIFO they should be discarded) and finally it can mark this position as available again. We have only one problem how to mark a file descriptor available using an array? One idea can be to use -1 as `fs_file_id` to identify a position that is marked as available (so we will need to set them to -1 when the vfs is initialized).
+
+In our case where we have no FIFO (++++ @DT: i think when the IPC chapter will be done we can spend few words on it here? ++++) or data-pipes, we can outline our close function as the following:
```
DeanoBurrito, line 338: Sounds good!

07_VirtualFileSystem/02_VirtualFileSystem.md
```diff
@@ -0,0 +1,433 @@
...
+
+The pseudo code above should give us an idea of what is the workflow of opening a file from a VFS point of view, as you can see the process is pretty simple in principle: getting the mountpoint_id from the vfs, if one has been found get strip out the mountpoint path from the path name, and call the fs driver open function, if this function call is succesfull is time to initialize a new vfs file descriptor item. 
+
+The `close` with signature: 
+
+```c 
+int close(int fildes);
+```
+
+will take a vfs file descriptor as parameter, and will search for it in the opened files list (using an array it will be found at `vfs_opened_files[fildes]`) and if found it should first call the fs driver function to close a file (if present), emptying all data structures associated to that file descriptor (i.e. if there are data on pipes or FIFO they should be discarded) and finally it can mark this position as available again. We have only one problem how to mark a file descriptor available using an array? One idea can be to use -1 as `fs_file_id` to identify a position that is marked as available (so we will need to set them to -1 when the vfs is initialized).
```
DeanoBurrito, line 336: This feels a bit janky starting a sentence like this. Maybe move the code above the paragraph, and then reference it by starting with something like "The example close() function shown above will take a vfs file ..."

07_VirtualFileSystem/02_VirtualFileSystem.md
```diff
@@ -0,0 +1,434 @@
...
+
+```C
+int fd;
+char *buffer = (char *) calloc(11, sizeof(char));
+int file_descriptor = open("/path/to/file/to_open", O_RDONLY);
+int sz = read(file_descriptor, buffer, 10) 
+buffer[sz] = '\0';
+printf("%s", buffer);
+close(file_pointer);
+```
```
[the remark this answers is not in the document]
dreamos82: thanks

07_VirtualFileSystem/02_VirtualFileSystem.md
```diff
@@ -0,0 +1,434 @@
...
+
+So once our open function has found the mountpoint for the requested file, eventually a new file descriptor item will be created and filled, and an id value returned. This id is different from the ine in the data structure, since it represent the internal fs descriptor id, while this one represent the vfs descriptor id. In our case the descriptor list is implemented again using an array, so the id returned will be the array position where the descriptor is being filled. 
+
+Why "eventually" ? Having found the mountpoint id for the file doesn't mean that the file exists on that fs, the only thing that exist so far is the mountpoint, but after that the VFS can't really know if the file exists or not, it has to devolve this task to the fs driver, hence it will call the implementation of a function that open a file on that FS that will do the search and return the an error if the file doesn't exists.
```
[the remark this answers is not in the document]
dreamos82: lol i don't even know how i came with that word XD

PR #36, opened 2023-01-05
Tar fs chapter

dreamos82:


07_VirtualFileSystem/03_TarFileSystem.md
```diff
@@ -0,0 +1,158 @@
+# The Tar File System
+
+In the previous chapter we have implemented the VFS layer. that will provide us with a common interface to access files and directories across different file systems on different devices. In this chapter we are going to implement our first file system and try to access it's content using the VFS function. As already anticipated we will implement the (US)TAR format.
```
DeanoBurrito, line 3: that -> That (new sentence starts with capital).

I would also maybe not say first file system, but first file system driver.

07_VirtualFileSystem/03_TarFileSystem.md
```diff
@@ -0,0 +1,158 @@
...
+| '5'   | Directory |
+| '6'   | Named Pipe |
+
+The _name of linked file_ field refers to symbolic links in the unix world, when a file is a link to another file, that field contains the value of the target file of the link.
+
+The USTar indictator (containing the string `ustar` followed by NULL), and the version field are used to identify the format being used, and the version field value is "00". 
+
+The `filename prefix` field, present only in the `ustar`, this format allows for longer file names, but it is splitted into two parts the `file name` field ( 100 bytes) and the `filename prefix` field (155 bytes)
+
+The other fields are either self-explanatory (like uid/gid) or can be left as 0 (TO BE CHECKED) the only one that needs more explanation is the `file size` field because it is expressed  as an octal number encoded in ASCII. This means we need to convert an ascii octal into a decimal integer. Just to remind, an `octal` number is a number represetend in base 8, we can use digits from 0 to 7 to represent it (is like binaries that use only 0 and 1 or hexadecimals that use from 0 to F). So for example:
```
DeanoBurrito, line 62: the last part is a bit clunky, maybe:
`similar to how binary (base 2) only have 0 and 1, and hexadecimal (base 16) has 0 to F.

07_VirtualFileSystem/03_TarFileSystem.md
```diff
@@ -0,0 +1,158 @@
...
+
+```c
+037 = 3 * 8^1 + 7 * 8^0 = 31
+```
+
+Remember we ignore the first 0 because it tells C that it is an octal number (and also it doesn't add any value to the final result!), since we are writing an os implementing this function should be pretty easy, so this will be left as exercise, we will just assume that we have the following function available to convert octal ascii to decimal: 
+
+
+```c
+int octascii_to_dec(char *number, int size);
```
DeanoBurrito, line 83: I would use `size_t` instead of `int` for size, makes more sense.

07_VirtualFileSystem/03_TarFileSystem.md
```diff
@@ -0,0 +1,158 @@
...
+```
+
+The size parameter tells us how many bytes is the digit long, and in the case of a tar object record the size is fixed: 12 bytes. Since we just need to implement a data structure for the header, this is left as exercise. Let's assume just that a new type is defined with the name `tar_record`.
+
+
+### Searching for a file
+
+Since the tar format doesn't have any file table or linked lists, or similar to search for files, we need everytime to start from the first record and scan one after each other, if the record is found we will return the pointer to it, otherwise we will eventually reach the end of the archive (file system in our case) meaning that the file searched is not present.
```
DeanoBurrito, line 91: I would mention that in your (the readers) design they may want to create a faster structure to lookup file locations. Maybe a linked list with each entry containing the filename, size and a pointer to where it's data starts in the tar image. That would drastically speed up searching.

DeanoBurrito: of course we dont have to implement it, but it might to hint to the reader

07_VirtualFileSystem/03_TarFileSystem.md
```diff
@@ -0,0 +1,158 @@
...
+| 100 | 8 	| File mode (octal) |
+| 108 | 8 	| Owner's numeric user ID (octal) |
+| 116 | 8 	| Group's numeric user ID (octal) |
+| 124 | 12 	| File size in bytes (octal) |
+| 136 | 12 	| Last modification time in numeric Unix time format (octal) |
+| 148 | 8 	| Checksum for header record |
+| 156 | 1 	| Type flag |
+| 157 | 100 | Name of linked file |
+| 57  | 6 	| UStar indicator, "ustar", then NULL |
+| 263 | 2 	| UStar version, "00" |
```
DeanoBurrito, line 37: I would add a note indicating that this is a string, just to be clear.

07_VirtualFileSystem/03_TarFileSystem.md
```diff
@@ -0,0 +1,158 @@
...
+* The filename length is less than 100 bytes, in this case it is stored into the `file_name` field
+* The length is greater than 100 bytes (up to 256) so in this case the filename is split in two parts, the first 100 bytes goes into the `file_name` field, the rest goes into the `filename_prefix` field. 
+
+An easy solution is to check first the searched filename length, if it less than 100 characters, so we can use just the `file_name` field, otherwise we can merge the two fields and compare than with the searched filename. The updated loop pseudo-code should look similar to this: , 
+
+```c
+
+char tar_filename[256];
+int zero_counter = 0;
+//The starting address should be known somehow to the OS)
```
DeanoBurrito, line 127: trailing bracket?

07_VirtualFileSystem/03_TarFileSystem.md
```diff
@@ -0,0 +1,158 @@
...
+tar_record current_record = tar_fs_start_address; 
+while (zero_counter < 2) {
+    if (is_zeroed(current_record) ) {
+        zero_counter++;
+        continue;
+    }
+    zero_counter = 0;
+    if ( tar_record.filename_prefix[0] != 0) {
```
DeanoBurrito, line 135: tar_record is the type! not variable name, you've used this a few times down below too.

07_VirtualFileSystem/03_TarFileSystem.md
```diff
@@ -0,0 +1,158 @@
...
+* The length is greater than 100 bytes (up to 256) so in this case the filename is split in two parts, the first 100 bytes goes into the `file_name` field, the rest goes into the `filename_prefix` field. 
+
+An easy solution is to check first the searched filename length, if it less than 100 characters, so we can use just the `file_name` field, otherwise we can merge the two fields and compare than with the searched filename. The updated loop pseudo-code should look similar to this: , 
+
+```c
+
+char tar_filename[256];
+int zero_counter = 0;
+//The starting address should be known somehow to the OS)
+tar_record current_record = tar_fs_start_address;
```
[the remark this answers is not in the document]
DeanoBurrito: which would mean updating the code use pointer notation too (`.` becomoes `->`)

DeanoBurrito (changes requested):

DeanoBurrito (approved):

PR #39, opened 2023-01-19
some more additions to the scheduling page

xyve7:
changes:
the code samples didn't dereference the context pointer when assigning it in the array of processes.
added a troubleshooting section and added a tip to make sure the rflags/eflags register is correctly set, as incorrectly setting it can cause the computer to disable interrupts and stop the timer.

### DIFF
--- a/04_Scheduling/02_Scheduler.md
+++ b/04_Scheduling/02_Scheduler.md
@@ -141,7 +141,7 @@ In order for this to happen, we need to modify our `schedule()` function a littl
 
 ```c
 cpu_status_t* schedule(cpu_status_t* context) {
-    processes_list[current_process_idx]->context = context;
+    processes_list[current_process_idx]->context = *context;
 
     do {
         current_process_idx = (current_process_idx++) % MAX_PROCESSES;
@@ -179,7 +179,7 @@ We'll modify our selection algorithm to take these new states into account:
 
 ```c
 cpu_status_t* schedule(cpu_status_t* context) {
-    processes_list[current_process_idx]->context = context;
+    processes_list[current_process_idx]->context = *context;
     processes_list[current_process_idx]->status = READY;
 
     do {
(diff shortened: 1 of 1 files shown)


04_Scheduling/02_Scheduler.md
```diff
@@ -221,6 +221,11 @@ void idle_main(void* arg) {
 
 The idle task is scheduled a little differently: it should only run when there is nothing else to run. You wouldn't want it to run when there is real work to do, because it's essentially throwing away a full quantum that could be used by another thread.
 
+## Troubleshooting
```
dreamos82, line 224: since i want to use pandoc to generate a pdf version of the notes when it is done, it is not perfect in handling titles not separated by a new line, so make sure to leave an empty line after each title.

xyve7: alright, ill fix that asap

DeanoBurrito:
Hey, first of all thanks for the PR!
You've identified a problem but I think the solution is incorrect: I believe intent was for the `process_t` struct to be defined as containing `cpu_status_t*` instead of `cpu_status_t`. By dereferencing (like in this PR) we'd be copying all ~184 bytes of the context into the new struct and we're now left with two copies (original one pushed on the stack by the interrupt handler, and the one we copied). That would probably be okay for now, but I think it'd be better to update the struct to container a pointer, since that can just point to the existing saved registers on the stack.

Also if you look down at [line 193](https://github.com/dreamos82/Osdev-Notes/pull/39/files#diff-1745701869ea7b309b74fa1ad166dc88f8f441880973a2f1348c3b6306efe9f3R193) you can see we return a pointer to the context, which we then use as the stack pointer when returning from the interrupt.

Troubleshooting tip is good! My only feedback would be to capitalize the title: `Interrupts Stop After Context Switch`, and that interrupts are disabled for the cpu not the whole computer.

dreamos82:
hi @xyve7 are you able to make the changes requested by @DeanoBurrito ?

xyve7:
> hi @xyve7 are you able to make the changes requested by @DeanoBurrito ?

i'll add the changes soon

DeanoBurrito:
Hey all, just bumping this. @xyve7, we'd love to get this fix in under your name (since you discovered the issue), would you mind making the requested changes?

xyve7:
@DeanoBurrito @dreamos82  sorry about the insane dely. i had this on my list of things to do but it slipped out of mind, so i apologize for pushing a change and then completely forgetting about it.

DeanoBurrito:
haha all good it happens, those TODO lists never get any shorter. Thanks again for the fixup :)

PR #43, opened 2023-03-28
Rearrange chapters (refactoring part 2)

dreamos82:


02_Architecture/02_Getting_Some_Output.md
```diff
@@ -0,0 +1,113 @@
+# Serial Logging (aka Hello World)
+
+During the developmen of our kernel we will need to debug a lot, and checking a lot of values, but so far our kernel is not capable of doing anything, and having proper video output with scrolling, fonts etc, can take some time to implement (and probably we need to check again some values), so we need a quick way of getting some output from our kernel, not necessarily on the screen.
```
DeanoBurrito, line 3: developmen -> development.
This is also a long sentence!

02_Architecture/02_Getting_Some_Output.md
```diff
@@ -0,0 +1,113 @@
...
+extern inline void outportb (int portnum, unsigned char data)
+{
+  __asm__ __volatile__ ("outb %%al, %%dx" :: "a" (data),"d" (portnum));
+}
+
+```
+
+### Initialization
+
+The second part again is pretty simple, we just need to send few configuration command for initializing the serial communication, the code below is copied from https://wiki.osdev.org/Serial_Ports#Initialization:
```
DeanoBurrito, line 42: I would mention debugcon here too, since it requires no setup (less room for error) and is often faster than the emulated serial chip. To use it simply write to port 0xE9.

02_Architecture/02_Getting_Some_Output.md
```diff
@@ -0,0 +1,113 @@
...
+
+```C
+#define PORT 0x3f8          // COM1
+ 
+static int init_serial() {
+   outb(PORT + 1, 0x00);    // Disable all interrupts
```
DeanoBurrito, line 48: there is a lot of PORT + xyz here, maybe explain that we're accessing the different registers of the serial chip? Maybe mention that register 0 is the input/output buffer too.

02_Architecture/02_Getting_Some_Output.md
```diff
@@ -0,0 +1,113 @@
...
+```
+
+Notice that usually the com1 port is mapped to address: *0x3f8*. The function above is setting just default values for serial communication. 
+
+### Sending a string
+
+Last thing to do is to create functions to print string/numbers on the serial. The idea is pretty simple, the current functions we created are handling single bytes/char, what we want is to send strings, so a good idea is to start with a function like: 
+
+```c
+void log_to_serial (char *string) {
```
DeanoBurrito, line 77: I would make this a `const char*`

02_Architecture/02_Getting_Some_Output.md
```diff
@@ -0,0 +1,113 @@
...
+}
+```
+
+The input parameter for this function is a string, so what it will do is looping through the variable `string` and printing each character until the symbol `\0` (End Of String) is found.
+
+This is the first function that we want to implement. 
+
+### Printing Digits
+
+Once we are able to print strings is time to print digits. The basic idea is simple, we takeread every single digit that compose the number, and print the corresponding character, luckily enough the digits symbols are consecutive in the ascii map, so for example:
```
DeanoBurrito, line 88: takeread

02_Architecture/02_Getting_Some_Output.md
```diff
@@ -0,0 +1,113 @@
...
+Once we are able to print strings is time to print digits. The basic idea is simple, we takeread every single digit that compose the number, and print the corresponding character, luckily enough the digits symbols are consecutive in the ascii map, so for example: 
+
+```c
+'0' + 1 // will contain the symbol '1'
+'0' + 5 // will contain the symbol '5'
+```
+
+How to get the single digits will depend on what base we are using (the most common are base 8, 10 and 16), let's assume we want for now just print decimals (base 10). For doing this we will use a property of division by 10, that the remainder of any integer number divided by 10 is always the same as the least significant digit, this means that for example $1235/10=123.5$ and $1235 \mod 10=5$, remember that in C (and other programming languages) a division between integers will ignore any decimal digit, so this means that $1235=123$. And what if now we divide 123 by 10? yes we get 3 as remainder, below the full list of divisions for the number 1235:
+
+* $1235/10 = 123$ and $1235 \mod 10 = 5$
```
DeanoBurrito, line 97: this is very confusing to read, maybe there's a better way to represent this?

PR #69, opened 2023-08-29
Fix paging chapter

dreamos82:


04_Memory_Management/03_Paging.md
```diff
@@ -2,41 +2,46 @@
...
+Paging is a memory management scheme that introduces the concept of **_logical addresses_** (virtual address) and **_virtual memory_**. On x86_\* architectures this is achieved via hardware. Paging enables a layer of translation between virtual and physical addresses, and virtual and physical address spaces, as well as adding a few extra features (like access protection, priviledge level protection).
 
-Paging introduces a few new concepts that are explained below.
+It introduces a few new concepts that are explained below.
 
 ### Page
-A _page_ is a contiguous block of memory, of fixed size. On x86 a page is 0x1000 bytes. It represents the smallest unit the hardware memory management unit can work with, and therefore the smallest unit we can work with! Each entry in a page table describes one page.
+A _page_ is a contiguous block of memory, of fixed size. The size depends on what the architecture supports (usually more than one).  On _x86\_64_ a page minum size is 0x1000 bytes. It represents the smallest unit the hardware memory management unit can work with, and therefore the smallest unit we can work with! Each entry in a page table describes one page.
```
DeanoBurrito, line 10: this sounds a little awkward and contradictory, maybe try:
`A page is a contiguous block of memory, with the exact size depending on what the architecture supports. On x86_64 we have page sizes of 4K, 2M and optionally 1G. The smallest page size is also called a page frame as it represents the smallest unit the memory management unit can work with, and therefore the smallest unit..`

DeanoBurrito (changes requested):

PR #81, opened 2024-01-22
Typo fixes

dreamos82:


06_Userspace/05_Example_ABI.md
```diff
@@ -58,6 +58,18 @@ void memcpy(void* src, void* dest, size_t count) {
 }
 ```
 
-## Summary
+## Summary and Next Steps
 
 At this point we should be ready to go off and implement our own system call interface, and maybe even begin to expose some kernel functions to userspace. Always keep in mind that values (especially pointers) coming from userspace may contain anything, so we should verify them and their contents as much as possible before passing them deeper into the kernel.
+
+The problem at this point is what are the syscall we should start to implement, the answer is as usual that depends mostly on the design decisions, and what we want to let user process access, and what we want to support in the future, for example to support an existing libc, we should implement a set of required syscalls, but for now let' s see what will most likely be useful and could act as a starting point for implementing them:
```
DeanoBurrito, line 65: One long sentence! Try breaking it up:

Now the question is what syscalls should we start to implement? As per usual this depends on the design of your kernel, and what interface we want to export userland. If unsure, take a look at POSIX (and the linux extensions), but we can also go the custom route. We should also keep in mind whether we want to port an existing libc later on, as this will require standard syscalls. For now we'll want to start with the following:
- A way to pass log messages to the kernel, so it can print them for us.
- A way to map, unmap, and modify protections of virtual memory.
- A way to terminate the current thread, since the wrapper function used in the scheduler chapter only works for kernel threads.

DeanoBurrito (changes requested):

PR #83, opened 2024-01-29
Fix RSDPDescriptor Mistake

maxtyson123:
I notice that you say:
• Signature: Is an 8 byte string, that must contain: “RSDT PTR” P.S. The string is not null
terminated

Where it should be “RSD PTR "
[Ref: OSDev Wiki](https://wiki.osdev.org/RSDP#:~:text=This%208%2Dbyte%20string%20(not%20null%20terminated!)%20must%20contain%20%22RSD%20PTR%20%22.%20It%20stands%20on%20a%2016%2Dbyte%20boundary.)

image

### DIFF
--- a/02_Architecture/06_ACPITables.md
+++ b/02_Architecture/06_ACPITables.md
@@ -46,7 +46,7 @@ struct RSDPDescriptor {
 
 Where the fields are:
 
-* *Signature*: Is an 8 byte string, that must contain: "RSDT PTR " **P.S. The string is not null terminated**
+* *Signature*: Is an 8 byte string, that must contain: "RSD PTR " **P.S. The string is not null terminated**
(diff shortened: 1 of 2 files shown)


DeanoBurrito:
Lgtm, thanks for the fix! Feel free to add yourself to the list in `99_Appendices/I_Acknowledgements.md` if you want, you can push the change to this PR. If not let us and know we'll merge it. 

Btw it may be more valuable to quote the acpi spec (where the magic value is defined) as the source rather than a publicly editable wiki.

maxtyson123:
Added Acknowledgement, the updated refrence is here: https://uefi.org/sites/default/files/resources/ACPI_1.pdf#page=91

DeanoBurrito:
Cool! Just check your acknowledgement entry again, you only changed one of the profile links (one still links to the previous person on that list).

maxtyson123:
Good spotting, all done now

DeanoBurrito:
It's why we review PRs 😉
Thanks again for the fix, merged this one.

PR #89, opened 2024-06-02
Add missing bits for tables in Paging document

AnErrupTion:
Hello there! 👋🏻

While doing paging work on my OS, I've noticed that some bits for the PML4/PDPR/PD & Page Table were missing in the `03_Paging.md` file, notably a few available bits & the PK (Protection Key) value. This PR adds them alongside an explanation of what the PK is. It also fixes the formatting of the tables within the file, because why  not! 🙃

Source, for reference: https://wiki.osdev.org/images/6/6b/64-bit_page_tables2.png

### DIFF
--- a/04_Memory_Management/03_Paging.md
+++ b/04_Memory_Management/03_Paging.md
@@ -153,9 +153,9 @@ Setting those bits must be done only once at early stages of boot process (proba
 
 PML4 and PDPR entry structures are identical, while the PD one has few differences. Let's begin by looking at the structure of the first two types:
 
-| 63   | 62 ... 59 | 58 ... 52 | 51 ... 40            | 39 ... 12              | 11  ...  9 |
-|------|-----------|-----------|----------------------|------------------------|------------|
-|**XD**| **PK**    | Available | _Reserved must be 0_ | **Table base address** | Available  |
+| 63   | 62 ... 59              | 58 ... 52 | 51 ... 40            | 39 ... 12              | 11  ...  9 |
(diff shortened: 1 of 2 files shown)


dreamos82:
Thanks for the contribution @AnErrupTion , it looks good to me, and i know why this bit was missing! 
Because i was using as reference a printed version of the intel manuals book pretty old, so at that time it wasn't present yet (in fact it was marked as avail to me). 

I'm ready to merge, but at this point i wonder do we want to speicify that this part could be not present in older cpus and we should make sure that is present first? What do you think guys @AnErrupTion @DeanoBurrito ?

AnErrupTion:
@dreamos82 Ah yes, I do think specifying these bits are either what they are currently or available is better (I didn't know they were all available on older CPUs).

DeanoBurrito:
According to the intel SDM if the protection key features aren't enabled (or supported) those bits are ignored (which was the behavior before these features were added). So its safe to write to them without checking for support, even if thats a bit silly because you'd end up with a GPF later-on when trying to set them in cr4.

I feel like your addition already implies this, but I agree with what @dreamos82 said :)

DeanoBurrito:
thanks for contribution btw, and feel free to add yourself to the list in `99_Appendices/I_Acknowledgements.md`

AnErrupTion:
@dreamos82 @DeanoBurrito Everything should be good to go now! Could you please check and confirm if that's the case?

DeanoBurrito:
Thanks, lgtm.

PR #100, opened 2024-10-10
Link Fix & Qemu Monitor Over Tellnet

maxtyson123:
Whilst using the PDF version of the book I noticed you could not access the website: https://jmnl.xyz/windowmanager/ due to a typo.

I also added information on how to access the monitor using Telnet as I found that to be easier to use with my setup and thought it may be helpful. 

(I use WSL but use qemu on the windows side, however my IDE creates the terminals in WSL so telnet helps bridge the two to access the monitor, additionally I use serial of studio so cant use monitor there)

### DIFF
--- a/99_Appendices/F_Debugging.md
+++ b/99_Appendices/F_Debugging.md
@@ -207,21 +207,28 @@ qemu-system-i386 [..other params..] -monitor unix:qemu-monitor-socket,server,now
 ```
 
 then on another shell, on the same folder where we started the emulator launch the following command:
-
 ```bash
 socat -,echo=0,icanon=0 unix-connect:qemu-monitor-socket
 ```
 
-This will prompt with a shell similar to the following:
 
+Another method is to use telnet to start the monitor. This is best for easier, cross-platform or remote use (albeit less secure).
 ```bash
-username@host:~/yourpojectpath/$ socat -,echo=0,icanon=0 unix-connect:qemu-monitor-socket
-QEMU 6.1.0 monitor - type 'help' for more information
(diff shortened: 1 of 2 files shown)


DeanoBurrito (changes requested):

99_Appendices/F_Debugging.md
```diff
@@ -212,19 +212,30 @@ then on another shell, on the same folder where we started the emulator launch t
...
 
-This will prompt with a shell similar to the following:
+
+Another method is to use telnet to start the monitor. This is handy for easier, cross-platform or remote use (albeit less secure).
+
+```bash
+qemu-system-i386 [..other params..] -monitor telnet::45454,server,nowait
+```
+
+This enables the monitor to listen on a specified port (ie, 45454). You can then connect to the QEMU monitor from another terminal or a remote machine (with networking setup) using Telnet:
```
dreamos82, line 222: i see telnet here with capital T, and in the previous sentence with lowercase t, maybe we want to use the same capitalization on both? Or is there a reason?

dreamos82 (commented):
Just a minor comment, anyway feel free to add yourself in the Acknowledgments appendix! :)

99_Appendices/F_Debugging.md
maxtyson123: Thank you, also fixed some to do with Qemu, didn't know if there were more

DeanoBurrito (approved):

DeanoBurrito:
lgtm, thanks for the fixes!

PR #101, opened 2024-10-10
Stivale replacement

dreamos82:


01_Build_Process/05_Generating_Iso.md
```diff
@@ -65,20 +65,24 @@ That's a lot of flags! Because of how flexible the iso format is, we need to be
 Now we have an iso with our kernel, config and bootloader files on it. We just need to install limine into the boot partitions, we can do this using `limine-deploy` like so:
 
 ```sh
-limine-deploy my_iso.iso
+./limine/limine bios-install $(IMAGE_NAME).iso
```
DeanoBurrito, line 68: you use `$(IMAGE_NAME)` here but I dont think its defined, seems I previously used `my_iso.iso` as the filename, can we use that for consistency?

01_Build_Process/05_Generating_Iso.md
```diff
@@ -49,7 +49,7 @@ The process for limine is a little more manual: we must build the iso ourselves,
 
 To get started we'll want to create a working directory to use as the root of our iso. In this case we'll use `disk/`. Next we'll need to clone the latest binary branch of limine (using `git clone https://github.com/limine-bootloader/limine.git --branch=v3.0-branch-binary --depth=1`) and copy 'limine.sys', 'limine-cd.bin', and 'limine-cd-efi.bin' into `disk/limine/`, creating that directory if it doesn't exist.
```
DeanoBurrito, line 50: lol this is showing the age of this part of the book: can you also update the branch name to `v8.x-binary` instead of `v3.0-branch-binary`?

DeanoBurrito: maybe we mention that this is the latest at the time of writing, but the reader should check the git repo themselves.

99_Appendices/J_Updates.md
```diff
@@ -58,3 +58,10 @@ Fifth book release
 * Added missing flags on page table/dirs entries
 * Rewrite and rearrangement of syscall/sysret section
 
+### Revision 5
+
+Relase date: TBD
+Sixth Book Release
+
+* _Stivale 2_ protocol sections have been replaced with Limine protocl, since _stivale2_ has been deprecated.
```
DeanoBurrito, line 66: protocl -> protocol

01_Build_Process/02_Boot_Protocols.md
```diff
@@ -203,115 +203,125 @@ The interface between a higher level language like C and assembly (or another hi
...
+
+__attribute__((used, section(".requests_end_marker")))
+static volatile LIMINE_REQUESTS_END_MARKER;
 ```
 
-The `framebuffer_*` fields can be used to ask for a specific kind of framebuffer, but leaving them to zero tells the bootloader we want to best possible available. The `next` field can be used to point to the next header tag, if we had another one we wanted. The full list of tags is available in the stivale2 specification (see the useful links appendix).
+Again they can be placed anywhere in the code, since their position will be decided by the section they belongs to.
 
-The last detail is to change the signature of our kernel entry function to:
+The last detail is to add the kernel start function (declared in the `ENTRY()` section in the linker script):
```
DeanoBurrito, line 310: since you mention below that the entry function doesnt require any particular signature - I think we can remove this sentence and the code snippet below?

01_Build_Process/02_Boot_Protocols.md
```diff
@@ -338,27 +348,17 @@ void* multiboot2_find_tag(uint32_t type)
...
-    return NULL;
+```c
+//somewhere in the code
+if ( framebuffer_request->response != NULL) {
+    // Do something with the response
 }
 ```
 
-The above function can be used with the defines in `stivale2.h`, which follow the format `STIVALE2_STRUCT_TAG_xyz_ID`, where `xyz` represents the feature that is described in the tag. For example, the framebuffer would be `STIVALE2_STRUCT_TAG_FRAMEBUFFER_ID` and the memory map is `STIVALE2_STRUCT_TAG_MEMMAP_ID`. It's a little verbose, but easy to search for.
+Is important to note, for every type of request the `response` field have a different type, in this case it is a pointer to a `struct limine_framebuffer_response`, for more info on all the available requests, and repsonses refer to the protocl documentation.
```
DeanoBurrito, line 362: protocl -> protocol

01_Build_Process/02_Boot_Protocols.md
```diff
@@ -338,27 +348,17 @@ void* multiboot2_find_tag(uint32_t type)
...
+
+The `framebuffer_*` fields can be used to ask for a specific kind of framebuffer, but leaving them to zero tells the bootloader we want to best possible available. The `next` field can be used to point to the next header tag, if we had another one we wanted. The full list of tags is available in the stivale2 specification (see the useful links appendix).
```
DeanoBurrito, line 364: this is leftover from stivale2, we should just delete this I think.

01_Build_Process/02_Boot_Protocols.md
```diff
@@ -203,115 +203,121 @@ The interface between a higher level language like C and assembly (or another hi
...
+static volatile LIMINE_REQUESTS_START_MARKER;
+
+__attribute__((used, section(".requests_end_marker")))
+static volatile LIMINE_REQUESTS_END_MARKER;
 ```
 
-This struct points to a list of tags, each containing details about the machine we're booted on. These are called struct tags (bootloader -> kernel) as opposed to the tags we defined before (header tags: kernel -> bootloader). To get info about a specific feature, simply walk the linked list of tags, the next tag's address is available in the `tag->next` field. The end of the list is indicated by a nullptr.
+Again they can be placed anywhere in the code, since their position will be decided by the section they belongs to.
+
+The last detail is to add the kernel start function (declared in the `ENTRY()` section in the linker script):
```
DeanoBurrito, line 310: I  think we remove this sentence too, since it referred to the block of code that is now gone :)

DeanoBurrito:
lgtm

PR #108, opened 2025-01-03
Update 08_Timers.md

mrjbom:


DeanoBurrito:
Hey thanks again for the PR - whats the reason behind this change? The LAPIC timer in TSC deadline mode only provides one shot operation.

mrjbom:
> Hey thanks again for the PR - whats the reason behind this change? The LAPIC timer in TSC deadline mode only provides one shot operation.

Hi. TSC deadline support is not guaranteed and may not be available.

DeanoBurrito:
It is optional, yes. All the optional timers mention this (TSC included) in their respective sections, alongside how to discover them. I'm not against putting a sentence up front that says "hey, some of these wont exist": my issue with the current wording is that it implies a periodic tsc deadline mode may be supported. 
If this sentence were to be added, it also applies to the HPET (and annoyingly the PIT now too, intel seemingly dropped it from arrowlake!).

mrjbom:
You're right. I suppose you shouldn't try to describe things in too much detail as in the spec. In general, you can close PR and not worry about it.

DeanoBurrito:
Its a hard balance to strike with regards to detail (and I'm not saying that we've got it dialed-in correctly), but I try to be mindful about not dumping too much information up front to keep things easy for the reader.

I'll close the PR for now, but like I mentioned before I'm happy if you want to write a sentence before the list of timers that points out that some of these are option and the reader should check their relevant section for how to determine if they exist. If you change your mind feel free to comment here or open a new PR :)

PR #111, opened 2025-02-12
Update I_Acknowledgments.md for username change

IAmTheNerdNextDoor:
Apologies for the inconvenient change and especially after a new version was put to print, but I recently changed my username across numerous platforms to all be the same (or relatively the same because some handles couldn't be acquired so I had to settle with a slightly altered version).

Is it okay to merge this? (Sorry once again!)


dreamos82:
Hey, no problem, sure you can change your username. 

Yeah, the timing is not the best, so you missed the latest printed version update of couple of weeks. But don't worry, your new username will be on the next book update!!

IAmTheNerdNextDoor:
Ah perfect! Thank you so much!!! I had originally put off changing my username here because my username was in this and would cause confusion, but I did it on an impulse earlier and completely forgot.

Thanks!

PR #114, opened 2025-03-25
IPC Implementation Rewrite

### LINKED ISSUE #113: IPC Messsage Passing
"If you’re performing this IPC as part of a system call from userspace, the memory containing the original
message is unlikely to be mapped in the receiver’s address space anyway, so we have to copy it into the kernel’s
address space, which is mapped in both processes." PG 226

But it is my understanding that whilst in usermode (CPL = 3) the process cannot directly access memory in the higher half addresses (e.g., 0xFFFFFFFF80000000 and above) because those addresses are typically reserved for the kernel and mapped with page table permissions that prevent user-space access.

TODO:
- Discuss
- Find reference in Intel Manual
- PR with proposed changes

maxtyson123:
resolves #113

As discussed in Issue #113 the current IPC message passing tutorial may be hard to follow for some users and the overall design could be reworked to be better. 

This PR proposes a improved design:
- Requires only 2 syscalls instead of 3
- Prevents a userspace program dictating when the buffer can be freed (see 36.1 How It Works - final bullet point)
- Allows for multiple messages
- More freedom for the process that owns the endpoint to handle messages how it deems fit for it's usecase

(This IPC mecanhisim was a combination of the current one in the book and my implementation in [Max OS](https://github.com/maxtyson123/MaxOS/tree/Processes)

### DIFF
--- a/02_Architecture/05_InterruptHandling.md
+++ b/02_Architecture/05_InterruptHandling.md
@@ -169,11 +169,11 @@ push %rbx
 push %r14
 push %r15
 
-call interrupt_disaptch
+call interrupt_dispatch
 
 pop %r15
 pop %r14
-//push other registers here
+//pop other registers here
 pop %rbx
 pop %rax
 
--- a/99_Appendices/I_Acknowledgments.md
+++ b/99_Appendices/I_Acknowledgments.md
@@ -26,3 +26,4 @@ In no particular order:
 - @vasilisalmpanis ([https://github.com/vasilisalmpanis](https://github.com/vasilisalmpanis))
 - @AntoninRuan ([https://github.com/AntoninRuan](https://github.com/AntoninRuan))
 - @sudw1n ([https://github.com/sudw1n](https://github.com/sudw1n))
+- @alexdev404 ([https://github.com/AlexDev404](https://github.com/AlexDev404))


DeanoBurrito:
Thanks for this, I'll take a look later today.

07_IPC/03_Message_Passing.md
```diff
@@ -4,25 +4,25 @@ Compared to shared memory, message passing is slightly more complex but does off
...
 
 You may wish to use the VFS and file descriptors in your own design, or something completely different! Several of the ideas discussed in the previous chapter on shared memory apply here as well, like access control. We won't go over those again, but they're worth keeping in mind here.
 
 ## How It Works
 
 After the initial setup, the implementation of message passing is similar to a relay race:
 
-- Process 1 wants to receive incoming messages on an endpoint, so it calls a function telling the kernel it's ready. This function will only return once a flag has been set on the endpoint that a message is ready, and otherwise blocks the thread. We'll call this function `ipc_receive()`.
-- Some time later process 2 wants to send a message, so it allocates a buffer and writes some data there.
+- Process 1  wants to receive incoming messages on an endpoint, so it calls a function telling the kernel to create an endpoint in our IPC manager. This function will setup and return a block of (userspace) memory containing a message queue. We'll call this function `create_endpoint()`.
```
inuyasha82, line 19: I think it could be useful to use italic for _Process 1_, etc. in this bullet point list.

07_IPC/03_Message_Passing.md
```diff
@@ -4,25 +4,25 @@ Compared to shared memory, message passing is slightly more complex but does off
...
+- Process 2 wants to send a message sometime later, so it allocates a buffer and writes some data there.
 - Process 2 now calls a function to tell the kernel it wants to send this buffer as a message to an endpoint. We'll call this function `ipc_send()`.
-- Inside `ipc_send()` the buffer is copied into kernel memory. In our example we'll use the heap for this memory. We can then set a flag on the endpoint telling it that a message has been received.
+- Inside `ipc_send()` the buffer is copied into kernel memory. In our example, we'll use the heap for this memory. We can then switch to process 1's address space and copy the buffer on the heap into the queue.
 - At this point `ipc_send()` can return, and process 2 can continue on as per normal.
-- The next time process 1 runs, `ipc_receive()` will see that the flag has been set, and copy the message from the kernel buffer into a buffer for the program.
-- The `ipc_receive()` function can also free the kernel buffer, before returning and letting process 1 continue as normal.
+- The message is now waiting at the end of the endpoint's message queue which processes 1 can do what it pleases with.
```
inuyasha82, line 24: processes 1? 
is either processes, or process 1?

07_IPC/03_Message_Passing.md
```diff
@@ -51,17 +62,22 @@ ipc_endpoint* first_endpoint = NULL;
 spinlock_t endpoints_lock;
 ```
 
-At this point we have all we need to implement a function to create a new endpoint. This doesn't need to be too complex, and just needs to create a new instance of our endpoint struct. Since we're using `NULL` to in the message buffer address to represent no message, we'll be sure to set that when creating a new endpoint. Also notice how we hold the lock when we're interacting with the list of endpoints, to prevent race conditions.
+At this point we have all we need to implement a function to create a new endpoint. This doesn't need to be too complex and just needs to create a new instance of our endpoint struct. Since we're using `NULL` in the message buffer address to represent no message, we'll be sure to set that when creating a new endpoint. Also notice how we hold the lock when we're interacting with the list of endpoints, to prevent race conditions. Note: `kmalloc()` assumes the use of kernel heap and `malloc()` assumes the active process's heap.
```
inuyasha82, line 65: use italic for _Note_. Just to separate it from the flow?

maxtyson123: Thank you, just added a commit with those fixes

dreamos82:
LGTM. 
Let's wait for @DeanoBurrito review.

dreamos82 (approved):

DeanoBurrito:
sorry for the delay, looking at this now. Time works differently where I'm from 😅

07_IPC/03_Message_Passing.md
```diff
@@ -4,25 +4,25 @@ Compared to shared memory, message passing is slightly more complex but does off
...
-- At this point `ipc_send()` can return, and process 2 can continue on as per normal.
-- The next time process 1 runs, `ipc_receive()` will see that the flag has been set, and copy the message from the kernel buffer into a buffer for the program.
-- The `ipc_receive()` function can also free the kernel buffer, before returning and letting process 1 continue as normal.
+- _Process 1_  wants to receive incoming messages on an endpoint, so it calls a function telling the kernel to create an endpoint in our IPC manager. This function will setup and return a block of (userspace) memory containing a message queue. We'll call this function `create_endpoint()`.
+- _Process 2_ wants to send a message sometime later, so it allocates a buffer and writes some data there.
+- _Process 2_ now calls a function to tell the kernel it wants to send this buffer as a message to an endpoint. We'll call this function `ipc_send()`.
+- Inside `ipc_send()` the buffer is copied into kernel memory. In our example, we'll use the heap for this memory. We can then switch to process 1's address space and copy the buffer on the heap into the queue.
```
DeanoBurrito, line 22: > copy the buffer on the heap into the queue.
is a little ambiguous, I think relying less on context would clean this up:
> copy the kernel's buffer into the endpoint's queue.

07_IPC/03_Message_Passing.md
```diff
@@ -33,16 +33,27 @@ As mentioned, there's some initial setup that goes into message passing: we need
...
 
 ```c
+
+typedef struct ipc_message{
+    void* message_buffer;
+    size_t message_size;
+    uintptr_t next_message;
+} ipc_message_t;
+
+ typedef struct ipc_message_queue{
```
DeanoBurrito, line 43: Is this type needed? since an endpoints and queues have a one-to-one relationship I dont think this is necessary. Instead embedding the list head in `ipc_endpoint` should be fine.

07_IPC/03_Message_Passing.md
```diff
@@ -33,16 +33,27 @@ As mentioned, there's some initial setup that goes into message passing: we need
...
+    ipc_message_t* messages;
+} ipc_message_queue_t;
+
 struct ipc_endpoint {
     char* name;
-    void* msg_buffer;
-    size_t msg_length;
+    ipc_message_queue_t* queue;
+    uint64_t owner_pid;
```
DeanoBurrito, line 50: I'd prefer `tid_t` over `uint64_t` here, it describes intent better. `uint64_t` may not always be a convenient type depending on the architecture.

07_IPC/03_Message_Passing.md
```diff
@@ -72,18 +89,22 @@ void create_endpoint(const char* name) {
...
-Now our endpoint has been added to the list! As always we omitted checking for errors, and we didn't check if an endpoint with this name already exists. In the real world you'll want to handle this things.
+Now our endpoint has been added to the list! As always we omitted checking for errors, and we didn't check if an endpoint with this name already exists. In the real world, you'll want to handle these things.
 
 ### Removing An Endpoint
 Removing an endpoint is also an important function to have. As this is a simple operation, implementing this is left as an exercise to the reader, but there are a few important things to consider:
 
-- What happens if there unread messages when destroying an endpoint? How do you handle them?
-- Who is allowed to remove an endpoint?
+- What happens if there are unread messages when destroying an endpoint? How do you handle them?
+- Who is allowed to remove an endpoint? (`owner_pid` would be useful here)
```
DeanoBurrito, line 107: what about:
> (hint: `owner_pid` is useful here)

07_IPC/03_Message_Passing.md
```diff
@@ -112,72 +133,83 @@ if (target == NULL)
...
-### Multiple Messages
+// Create the message
+ipc_message_t* new_message = (ipc_message_t*)malloc(sizeof(ipc_message_t));
+void* new_buffer = malloc(size);
+new_message -> message_buffer = memcpy(new_buffer, kernel_copy, size);
+new_message -> message_size = size;
+new_message -> next_message = 0;
 
-In theory this works, but we've overlooked one huge issue: what if there's already a message at the endpoint? You should handle this, and there's a couple of ways to go about it:
+// Add  the message to the queue
```
DeanoBurrito, line 163: extra space after `Add`

07_IPC/03_Message_Passing.md
```diff
@@ -112,72 +133,83 @@ if (target == NULL)
...
+// Left for the reader to do, trivial linked list appending
```
DeanoBurrito, line 164: this reads like a note I'd leave for myself rather than a hint. What about:
> // append message struct to the endpoint's queue, implementation is left as an exercise for the reader.

07_IPC/03_Message_Passing.md
```diff
@@ -112,72 +133,83 @@ if (target == NULL)
...
-- In this example we've allowed for messages of any size to be sent to an endpoint, but you may want to set a maximum message size for each endpoint when creating it. This makes it easier to receive messages as you know the maximum possible size the message can be, and can allocate a buffer without checking the size of the message. This might seem silly, but when receiving a message from userspace the program has to make a system call each time it wants the kernel to do something. Having a maximum size allows for one-less system call. Enforcing a maximum size for messages also has security benefits.
+- We've described a double-copy implementation here, but you might want to try a single-copy implementation. Single-copy implementations *can* be faster, but they require extra logic. For example, the kernel will need to access the recipient's address space from the sender's address space, how do you manage this? If you have all of the physical memory mapped somewhere (like an identity map, or direct map (HHDM)) you could use this, otherwise, you will need some way to access this memory.
```
DeanoBurrito, line 206: > , otherwise, 
Too many commas, the second one can be omitted I think.

07_IPC/03_Message_Passing.md
```diff
@@ -112,72 +133,83 @@ if (target == NULL)
...
+- A process waiting on an endpoint (to either send or receive a message) could be waiting quite a while in some circumstances. This is a time when the cpu could be doing work instead of blocking and spinning on a lock. A simple optimization would be to put the thread to sleep, and have it be woken up whenever the endpoint is updated: a new message is sent, or the current message is read.
+- In this example we've allowed for messages of any size to be sent to an endpoint, but you may want to set a maximum message size for each endpoint when creating it. This makes it easier to receive messages as you know the maximum possible size the message can be, and can allocate a buffer without checking the size of the message. This might seem silly, but when receiving a message from userspace the program has to make a system call each time it wants the kernel to do something. Having a maximum size allows for one less system call. Enforcing a maximum size for messages also has security benefits.
```
DeanoBurrito, line 208: the point about saving a system call is irrelevant for this implementation, since the message contents and metadata are already in the recipients address space. There's no system call in the example you give your receiving messages.
The rest is good.

DeanoBurrito (changes requested):
Writing is good overall, but there's a couple of design issues I see:

1) The kernel is accessing the userspace heap (see the distinction you made about `kmalloc()` vs `malloc()`)? This is a bad idea, as the kernel now relies on user code to 'do the right thing' and provide a working `malloc()` to allocate the recipients message buffers. Or alternatively the kernel has to be involved in userspace heap management (the user program might not even have a heap, or may have multiple, or some per-thread caches - how do you deal with that on the kernel side?).

This is a tricky one to deal with, the common approach is that userspace allocates the memory and informs the kernel of where the buffer is located as part of a system call. This is where a dedicated `ipc_receive()` function is handy, because the recipient can pass the buffer they wish the message to be copied into. You could also have a big block of memory attached to the endpoint and have that act as shared memory between the kernel and receiving process (this would need some synchronization if the receiving process isnt blocking on reads).

2) Similar to 1, this version of message parsing combines IPC concepts and userspace/kernel interactions into one chapter. In a real implementation there may be some overlap, but I think it muddies the waters when it comes to explaining something new. It would be better to break this down into smaller ideas: how to move messages between two address spaces (all of this happens in the kernel, this is what the original message parsing example did), and then the API presented to userspace (not previously covered).

dreamos82:
@maxtyson123 any news on this PR? Do you want to make the changes requested?

IAmTheNerdNextDoor:
> @maxtyson123 any news on this PR? Do you want to make the changes requested?

and he never responded...

maxtyson123:
Hi sorry, what would you like done? Quite busy right now but I can being working on it when I have spare time.

maxtyson123:
Just to be clear before I begin working on this:

1. Now that you mention it, I do agree that it wasn't the best practice, albeit for different reasons. In [Max Os](https://github.com/maxtyson123/MaxOS) each process has its own [Memory Manager](https://github.com/maxtyson123/MaxOS/blob/main/kernel/include/memory/memorymanagement.h) that handles that process's memory (Note: this may change when I work on actually implementing threads). I now realise that my suggested method "locks" the reader into my design of managing memory in the kernel and in that way. How would you suggest this is fixed, going back to the 'read()' design or something new?
2. It should be broken up into two sections. One subsection of the "Userspace" section for [resource management](https://github.com/dreamportdev/Osdev-Notes/issues/113#issuecomment-2750387205), detailing HOW to move information between address spaces. _Note: I haven't finished my filesystem implementation (what I'm working on for [Max Os](https://github.com/maxtyson123/MaxOS) now) so if we want to address files in there aswell (which I think mat be a good idea) I will need to finish that first._ Separately, in the IPC section, the code should be changed to build on top of this API.   

P.S. Is there a better place to talk with easier back and forth? That way we can keep the public github discussion to what is actually relevant to the PR and can also have a method smaller things

DeanoBurrito:
Hey Max, all good - life is like that sometimes. Glad to see you're still interested in this :)
We have a discord server we hang out in (I'm a little absent myself these days, but I do eventually get around to responding to messages). Here's the invite: https://discord.gg/X5YmgDKW

As for your questions:
1. That's pretty standard to have a 'memory manager' (in this case its your `VirtualMemoryManager` class) per address space, which is usually linked one-to-one with a process. A number of threads would share the address space of the process, so that all makes sense. I'm not sure about the purpose of your `MemoryManager` class though, it looks like it allocates memory for a heap in the address space it's attached to (so far so good), but never maps the pages as user-accessible, so you have a kernel heap in the higher half and then one per address space in the lower half if I understand correctly? This sounds like it's breaking the 'lower half is for userspace' idea. So my suggestion there is to remove the lower-half kernel heaps, keep kernel stuff in the higher half, and then yeah as you mentioned - going back to the `read()` design. Or something similar where userspace passes the buffer to the kernel. 

The other big benefit to the `read()` approach is the user thread can block until there's a new message, or rather you can see it as `read()` will only return when there's a new message to be processed (ignoring edge cases where the read times out, or is cancelled). If the user thread only wants to poll, you can support that to, and in that case it's like the section you wrote where it checks for a new message and if there's none it carries on with other work.

2. Yeah that division sounds good. It'd be good to focus on providing general theory first for exposing kernel resources to userspace (via handles/file descriptors/objects/your preferred terminology), and how you might go about moving data in and out of those resources. That framework could then easily be applied to files, ipc and other things. 
Personally I'd like to keep the focus of the userspace chapter on dealing with userspace, so I wouldnt object to some small examples of how this might be used - but I think it's better to keep file access related stuff to the vfs chapter.

PR #120, opened 2025-07-09
Update README.md

dreamos82:


sudw1n:
Not sure if I'm doing something wrong but that invite link seems to be invalid or expired.

PR #124, opened 2025-07-26
Fix typo on line 22 of Interrupt Handling section

AlexDev404:
This should fix a typo in the sentence where it says,

> "Don't let NMIs scare you, we've never **run** actually run..."

Instead of:

> "Don't let NMIs scare you, we've never **actually run**"


DeanoBurrito:
good find, thanks!

AlexDev404:
> good find, thanks!

Haha, no problem 🤣🤣
Glad I could help!!! 🙂

PR #127, opened 2025-09-02
Resource Magement

### LINKED ISSUE #113: IPC Messsage Passing

maxtyson123:
Here's the base write-up. Once settled on the design, I will go through and mix this in with the files and IPC. 

closes #113

### DIFF
--- a/08_VirtualFileSystem/02_VirtualFileSystem.md
+++ b/08_VirtualFileSystem/02_VirtualFileSystem.md
@@ -301,7 +301,7 @@ struct fs_operations_t {
 	ssize_t (*write)(int file_descriptor, const void* write_buffer, size_t nbyte);
 };
 
-typedef struct fs_operations_t fs_operations_t;
+typedef struct fs_operations_t;
 ```
 
 The basic idea is that once `mountpoint_id` has been found, the vfs will use the mountpoint item to call the fs driver implementation of the open function, remember that when calling the driver function, it cares only about the relative path with the mountpoint folder stripped, if the whole path is passed, we will most likely get an error. Since the fs root will start from within the mountpoint folder we need to get the relative path, we will use the `get_rel_path` function defined earlier in this chapter, and the pseudocode for the open function should look similar to the following:


06_Userspace/06_Resource_Management.md
```diff
@@ -0,0 +1,101 @@
...
+
+
+## Generic API
+Now that we have a way of representing resource, we to need define how a process can interact with them. Generally, having a different syscall for each resource type can lead to lots of repeated code segments and make the kernel interface harder to maintain and extend. Instead the kernel can expose a minmal and uniform API that every resource supports. The generic interface for a resource consists of four primary functions: `open`, `read`, `write`, and `close` and by restricting all resources to this same interface we can reduce the complexity of both the kernel and userspace. To begin the implementation of this, our `resource_t` needs extending with a table of function pointers to support these operations. Each resource can then provide it's own implementation of thse four functions whilst the generic interface remains the same.
```
DeanoBurrito, line 64: representing resource -> representing resources

DeanoBurrito: This is a bit wordy after the first sentence, it sounds as if you're unsure of what you're writing. Have some confidence! You could try something more like:

> Having a different set of syscalls for each type of resource leads to a lot of duplicated code, and would be a massive API for programmers to use and maintain. Instead we chose a set of operations common to all resources, and have them operate on our abstract resources. This generic interfaces consists of four functions: `open`, `read`, `write` and `close`. To begin our implementation of this ... (rest is fine I think).

06_Userspace/06_Resource_Management.md
```diff
@@ -0,0 +1,101 @@
...
+In this example `FILE` represents a file on the disk, `MESSAGE_ENDPOINT` is used for an IPC message queue and `SHARED_MEM` for a shared memory region between prcesses. As the kernel grows this struct can be extened to support more resource types. 
+
+Next, we need a generic representation of a resource inside the kernel. This can be defined by the `resource_t` struct:
+```
+typedef struct {
+    resource_type_t type;
+    void* impl;
+} resource_t;
+```
+The `type` field tells the kernel what kind of resource it is and the `impl` pointer allows the kernel to attach the resource specific implmentation of that resource. For example, a file's `impl` could point to a struct holding the file's offset and indoe or for shared memory it could point to the physical address of that region.
```
DeanoBurrito, line 25: implmentation -> implementation

06_Userspace/06_Resource_Management.md
```diff
@@ -0,0 +1,101 @@
...
+} resource_t;
+
+typedef struct resource_functions {
+    size_t (*read)(resource_t* res, void* buf, size_t len);
+    size_t (*write)(resource_t* res, const void* buf, size_t len);
+    void (*open)(resource_t* res);
+    void (*close)(resource_t* res);
+} resource_functions_t;
+```
+Here, `funcs` is the dispatch table that tells the kernel how to perform each operation for each resource. With this, each function pointer can be set differently dependin on wether the resource is a file, IPC endpoint or something else.
```
DeanoBurrito, line 79: dependin -> depending

06_Userspace/06_Resource_Management.md
```diff
@@ -0,0 +1,101 @@
...
+
+Operations are defined to be blocking by default, meaning that if a resource is not ready (for example, no data to read), the process is suspended until the operation can complete. Each resource type can override these generic operations to provide behavior specific to that resource. For example, a file resource can replace the write function with one that writes data to disk, while an IPC message resource could implement write to enqueue a message, allowing the same API call to behave differently depending on the resource.
```
DeanoBurrito, line 81: wrt the second sentence and beyond: this was all covered by the previous paragraph I feel, if you wanted to expand upon that idea I would maybe do it there.

06_Userspace/06_Resource_Management.md
```diff
@@ -0,0 +1,101 @@
...
+
+It has been left as an exercise to the reader to decide on how they want to handle extending this design for extra resource-specific functionality (ie, renaming a file). There are two (of many) ways to do this, each with its own trade-off. Firstly, a simpler design would be to just add more syscalls to handle this; however, this means the ABI grows as your kernel manages more resources. Another approach would be to pass an additional `size_t flags` parameter and let the resource-specific operation handle it, which would keep the original four operations but with added complexity.
```
DeanoBurrito, line 83: The flags param is nice, but not for this purpose. Stuff like renaming, creating files/endpoints is best handled by dedicated syscalls anyway, trying to force them into the 3 syscalls defined above would get messy very quickly. What we were talking about with the flags idea is reducing the need for something like `open()`, `open_readonly()` and so on (similar operations that have slight differences in behaviour).

06_Userspace/06_Resource_Management.md
```diff
@@ -0,0 +1,101 @@
...
+        return -1;
+
+    return res->funcs->read(res, buf, len);
+}
+```
+The other operations (`write`, `open`, `close`) would follow the same pattern above: get the resource from the handle and then call the appropriate function from the `funcs` table if supported. With this indirect approach, the kernel's syscall layer is kept minimal whilst allowing for each resource type to have its own specialised behavior.
+
+## Data Copying
+Another thing left as an exercise to the user is to decide their method of copying data between userspace and the kernel.
+One option is to use the userspace provided buffersm, which is efficient due to a single copy but does require sanitization of pointers and lengths to ensure safety. Some things to consider are other threads in the same address space modifying memory at the same address  Another option is to copy into a kernel buffer first, which simplifies the sanitization but has the added overhead and loss of performance.
```
DeanoBurrito, line 101: With using the user buffers, its not necessarily a single copy - you may be able to operate directly on the buffer (in which case its zero-copy). I think its worth driving home the danger of this approach a bit more: another user thread can write or try to unmap/map on top of the buffer it passed to the kernel. You cant really hold a lock protecting the user's address space for the entire operation on the kernel side (can you imagine the slowdown!) so the only real solution here is to have a try/catch setup helped by the page fault and GP fault handlers. The complexity here is well worth it, but its a dangerous thing if not done correctly.

There's also a missing fullstop before the last sentence. I also would qualify the loss of performance, its probably faster for smaller buffers, but yes potentially much slower in larger cases.

06_Userspace/06_Resource_Management.md
```diff
@@ -0,0 +1,118 @@
...
+
+To do this, we implement an API where every resource can be opened, read from, written to, and closed using the same syscalls. Through this design, the kernel is kept small, whilst also letting new resources be added in the future with minimal change to both kernel and user code.
+
+## Resource Abstractation
+
+When talking about _resources_, we need a way to distinguish between the different types that the kernel may expect to provide to userspace. Each resource behaves differently internally, but from the view of the userspace process, everything should be accessible from the same set of syscalls. In order to achieve this, we define an enum of resource types to allow the kernel to tag each resource with its category. This way, when a system call is made, the kernel knows how to dispatch the request.
```
dreamos82: Even if in  the example is an enum, I don't think that it has to be strictly it. 
The sentence is kind of assuming and telling it, I think that it can be made more generic. 

Something like: ` In order to achieve this we need to define a set of resource types to allow the kernel to tag each resources, for example we can use an enum... etc.`

06_Userspace/06_Resource_Management.md
```diff
@@ -0,0 +1,118 @@
...
+size_t pid;
+status_t process_status;
+
+// Other fields
+
+resource_t* resource_table[MAX_RESOURCES];
+} process_t;
+```
+
+Now each process has a resource table that is a map of integers, called _handles_, to the kernel resource objects. A handle is simply an identifier returned by the kernel when opening a resource that is later used by the user to inform what resource the operation should be performed upon. This indirection is important because we do not want to expose any kernel pointers directly to a userspace process. Even if they cannot be used there, passing them could still create security or stability risks. This way, the resource structure is not exposed to userspace. Because of this, the same handle number in different processes can refer to different resources. For example, in Unix, handles `0`, `1`, and `2` refer to stdio for each process and are called "file descriptors".
```
dreamos82: stdio and stderr? :D

06_Userspace/06_Resource_Management.md
```diff
@@ -0,0 +1,118 @@
...
+    size_t (*read)(resource_t* res, void* buf, size_t len);
+    size_t (*write)(resource_t* res, const void* buf, size_t len);
+    void (*open)(resource_t* res);
+    void (*close)(resource_t* res);
+} resource_functions_t;
+```
+
+Here, `funcs` is the dispatch table that tells the kernel how to perform each operation for each resource. With this, each function pointer can be set differently depending on whether the resource is a file, IPC endpoint, or something else. Operations are defined to be blocking by default, meaning that if a resource is not ready (for example, no data to read), the process is suspended until the operation can complete. Each resource type can override these generic operations to provide behavior specific to that resource.
+
+It has been left as an exercise to the reader to decide on how they want to handle extending this design for extra resource-specific functionality (ie, renaming a file). A simpler design may be to just add more syscalls to handle this; however, this means the ABI grows as your kernel manages more resources.
```
dreamos82: functionality of functionalities?

06_Userspace/06_Resource_Management.md
```diff
@@ -0,0 +1,118 @@
...
+
+Another thing left as an exercise to the user is to decide their method of copying data between userspace and the kernel.
+One option is to use the userspace provided buffers, which is efficient due to a single copy but does require sanitization of pointers and lengths to ensure safety. Some things to consider are other threads in the same address space modifying memory at the same address. Another option is to copy into a kernel buffer first, which simplifies the sanitization but has the added overhead and loss of performance. 
+
+With using the user buffers, it's not necessarily a single copy, and you may be able to operate directly on the buffer (in which case it's zero-copy). Although, this can be dangerous as another user thread can write to, unmap, or remap the buffer while the kernel is operating on it. Holding a lock over the address space for that process throughout the entire duration of the resource operation is impractical, so the kernel must instead rely on fault handling. By faulting when the process tries to access the memory that the kernel is working with, it allows this behaviour to e caught and the kernel can try to abort or retry the operation safely.
```
dreamos82: At end of sentence: `to e caught` (missing b)

08_VirtualFileSystem/02_VirtualFileSystem.md
```diff
@@ -362,19 +372,20 @@ mountpoint_t *get_mountpoint_by_id(size_t mountpoint_id);
...
 ```c
-ssize_t read(int fildes, void *buf, size_t nbyte);
+ssize_t read(resource_t* res, void *buf, size_t nbyte);
 ```
 
-Where the parameters are the opened file descriptor (`fildes) the buffer we want to read into (`buf`), and the number of bytes (`nbytes`) we want to read.
+Where the parameters are the opened file resource (`res`) the buffer we want to read into (`buf`), and the number of bytes (`nbytes`) we want to read.
 
-The read function will return the number of bytes read, and in case of failure -1. Like all other vfs functions, what the read will do is search for the file descriptor with id `fildes`, and if it exists call the fs driver function to read data from an opened file and fill the `buf` buffer.
+The read function will return the number of bytes read, and in case of failure -1. Like all other vfs functions, what the read will acutally do is verify the resource and then call the fs driver function to read data from an opened file and fill the `buf` buffer.
```
dreamos82: acutally?

08_VirtualFileSystem/02_VirtualFileSystem.md
```diff
@@ -386,28 +397,29 @@ And we have the following code:
...
-        int bytes_read = mountpoints->operations->read(fs_file_id, buf, nbytes);
-        if (opened_files[fildes].buf_read_pos + nbytes < opened_files[fildes].file_size) {
-            opened_files[fildes].buf_read_pos += nbytes;
-        } else {
-            opened_files[fildes].buf_read_pos = opened_files[fildes].file_size;
-        }
-        return bytes_read;
-    }
-    return -1;
+ssize_t read(resource_t* res, void *buf, size_t nbytes) {
```
dreamos82: Ok so this is read, and above we are using read_resource? And before we use again read. Let's decide on a name :D

dreamos82 (changes requested):
see previous comments

06_Userspace/06_Resource_Management.md
```diff
@@ -0,0 +1,118 @@
...
+
+  // Invalid handle
+  if (handle < 0 || handle >= MAX_HANDLES)
+      return NULL;
+  
+    return proc->table[handle];
+}
+```
+
+You would also want to implement two other functions: `int register_resource(process_t* proc, resource_t* rec)` that finds a free handle and stores the resource in the array and also `int remove_resource(process_t* proc, resource_t* rec)` that then marks that handle usable and frees the memory of that resource on clean up.
```
DeanoBurrito: No criticism here, I like the way you've structured this wrt to showing the bare minimum critical code and then describing what might be useful and leaving the reader to figure it out. Feel free to mark this comment as resolved.

06_Userspace/06_Resource_Management.md
```diff
@@ -0,0 +1,118 @@
...
+    size_t (*read)(resource_t* res, void* buf, size_t len);
+    size_t (*write)(resource_t* res, const void* buf, size_t len);
+    void (*open)(resource_t* res);
+    void (*close)(resource_t* res);
+} resource_functions_t;
+```
+
+Here, `funcs` is the dispatch table that tells the kernel how to perform each operation for each resource. With this, each function pointer can be set differently depending on whether the resource is a file, IPC endpoint, or something else. Operations are defined to be blocking by default, meaning that if a resource is not ready (for example, no data to read), the process is suspended until the operation can complete. Each resource type can override these generic operations to provide behavior specific to that resource.
+
+It has been left as an exercise to the reader to decide on how they want to handle extending this design for extra resource-specific operations (ie, renaming a file). A simpler design may be to just add more syscalls to handle this; however, this means the ABI grows as your kernel manages more resources.
```
DeanoBurrito: Fair, I would maybe mention `ioctl()` as an example of how this might look, gives the reader a jumping off point.

08_VirtualFileSystem/02_VirtualFileSystem.md
```diff
@@ -8,6 +8,8 @@ To keep our design simple, the features of our VFS driver will be:
 * No extra features like permissions, uid and gid (although we are going to add those fields, they will not be used).
 * The path length will be limited.
 
+This VFS is built on top of the previously defined resource API (see Resource Management). The VFS creates a file-specific `resource_t` for each open file and returns a handle that is then used with the resource manager's generic `read`, `write`, and `close` functions that will perform the proper _handle-to-resource_ lookup and call our VFS functions.
```
DeanoBurrito: I think this paints the wrong picture. The resource api and vfs are separate subsystems that interact but the vfs doesnt depend on the resource api, it can (and should) exist without it. The two would only actually interact when something higher level connects them (like a syscall request).

The resource subsystem makes little sense for in-kernel operations, as you can use raw pointers to things, so why incur the extra cost?

dreamos82:
@maxtyson123 when you have time please review the comments from @DeanoBurrito  (my plan is to make a new book release after this PR has been merged) :D

PR #128, opened 2025-09-30
Fix some typos in paging section.

frischerZucker:
This should fix some typos on line 201 of 03_Paging.md:
- it said "page teble" instead of "page table" 
- there was an "it" missing in "if it is set"
- split the first sentence into two parts, so that it is easier to read


frischerZucker:
Yeah sure, why not :)

dreamos82:
Thanks! @frischerZucker

PR #132, opened 2025-10-31
doc(interrupt): remove type casting to  uint64_t

realstealthninja:
The type taken by the function is void* but here we are passing uint64_t is this right?


dreamos82:
Thanks. Do you want to add yourself to the acknowledments list here https://github.com/dreamportdev/Osdev-Notes/blob/master/99_Appendices/I_Acknowledgments.md?

dreamos82:
Thanks for your PR! And thanks for your appreciation, it means a lot to us!!

PR #134, opened 2025-12-16
Shtifting bug explanation

dreamos82:


04_Memory_Management/05_Heap_Allocation.md
```diff
@@ -416,7 +416,7 @@ This means that the allocator (before marking this location as free and returnin
 
 ![The heap status after the merge](/Images/heap_example_after_merge.png)
 
-The fields in bold are the fields that are changed. The exact implementation of this code is left to the reader.
+The fields in bold are the fields that are changed. The exact implementation of this code is left to the reader. Please note that some books refers to this technique, calling it _coalescing_, but is the same concepts.
```
KnudAndersen: You should just introduce it as coalescing in the first place. It is already standard terminology in undergraduate curricula. So maybe early on, we could ask the question: what are some things we want to do with (free) heap blocks? The two natural answers are:
1) Splitting -> motivate with an example
2) Coalescing -> motivate with an example
The examples should show that splitting and coalescing go hand in hand. Imagine you only split, then your block sizes will be non-increasing, which means harder chance to repurpose free blocks. This is why coalescing is helpful.

dreamos82: I just reread the whole paragraph, and the main thing is that the approach is basicall incremental, so the "coalesce" concept is only introduced by consecutive examples expanding the previous implementation, and i haven´t found a better place of where to introduce that name earlier. I'll try anyway to rephrase the paragraph.

dreamos82: I tried to introduce the coalescing term at the beginning of the chapter. 

Also I think that the splitting and merging are explained in the way you suggested.

08_VirtualFileSystem/03_TarFileSystem.md
```diff
@@ -33,14 +33,15 @@ As anticipated above, the header structure is a fixed size struct of 512 bytes.
...
-| 57  | 6 	| UStar indicator, "ustar", then NULL |
+| 257  | 6 	| UStar indicator, "ustar", then NULL |
 | 263 | 2 	| UStar version, "00" (it is a string) |
 | 265 |	32 	| Owner user name |
 | 297 |	32 	| Owner group name |
 | 329 |	8 	| Device major number |
 | 337 |	8 	| Device minor number |
 | 345 |	155 | Filename prefix |
 
+The sum of all sizes, anyway is not 512 bytes, but 500, so the extra space is filled with zerosextra space is filled with _0s_.
```
DeanoBurrito: This could be reflowed to sound smoother: `While a tar file header occupies 512 bytes, currently only 500 bytes are defined. The extra space is filled with zeroes.`

99_Appendices/C_Language_Info.md
```diff
@@ -96,6 +96,38 @@ It is worth mentioning that inline assembly syntax is the At&t syntax, so the us
 asm("movl $5, %rcx;");
 ```
 
+## Dealing With Literals and Bitwise Operation
+
+This is one of the most misleading and subtle issue we can face while osdeving. And most of the time we face it the hard way.
```
DeanoBurrito: Content is fine overall, just some general comments:
- use capitals are the start of sentence and not starting with conjunctions (e.g. and).
- there's some baked-in assumptions about type sizes in C, `int` isnt necessarily 32-bits. The issue is that `int` is smaller than the fixed-size `uint64_t` type.
- it's overall quite wordy, the same example is given twice and the comparison to languages other than c/c++ isnt needed I think.
- the real solution is to ensure that the immediate value is of the correct type for the next operation, which you give examples of but dont actually say that.
- The introduction is quite ominous haha. Perhaps consider something like: 

> There are some subtle bugs that can be encountered when when using immediate values in C, due to operator precedence and integer promotion rules.

PR #140, opened 2026-02-07
Update 01_README.md

dreamos82:


00_Introduction/01_README.md
```diff
@@ -28,3 +28,5 @@ Below the list of parts that compose the book:
 * *Going Beyond* - The final part (for now). We have implemented all the core components of a kernel, and we are free to go from here. This final chapter contains some ideas for new components that we might want to add, or at least begin thinking about.
 
 In the appendices we cover various topic, from debugging tips, language specific information, troubleshooting, etc.
+
+The main purpose of this book is to explain the whole process, and we decided to keep things as simple as possible. So for more experienced users, some design decision could look over simplified, or not optimal, but this is intentional, the focus is on the how more than on the whole design.
```
DeanoBurrito: The content is good, I would actually suggest we put this earlier (as a paragraph between lines 3 and 5) and tweak the grammar:

> The main purpose of this book is to explore the whole process from a beginners point of view: we decided to keep things as simple as possible for this reason. Readers with more experience will notice that some design decisions and comments are not the optimal way to implement some things, or that some descriptions are over simplified - this is on purpose. The intent here is to provide a jumping-off point for the readers to continue on from.

